PR #31, opened 2015-01-21
check if the url provided matches the source associated with the token

deniak:


app.js
```diff
@@ -201,7 +201,8 @@ function orchestrate(spec, isManifest) {
     spec.jobs['token-checker'].status = 'pending';
     var shortlink; //@todo retrieve shortlink
     return TokenChecker.check(shortlink, token).then(function(authReport) {
-      if(authReport.authorized) {
+      var matchSource = spec.url.substring(0, authReport.source.length) === spec.url;
```
astorije: Shouldn't that check that `spec.url.substring(/* ... */) === authReport.source` instead? If not, I am not sure to understand what is being checked.

astorije:
@deniak, apart from the comment I left inline, I have a couple of questions for this:

- Should this check be done by echidna or by the publication system instead?
- If I understand your check properly, you check that the beginning of the URL of source to publish matches the registered URL. That would be a problem if someone registers something like `https://github.com`, no?

astorije:
By the way, if you rebase your commit with master and force-push, it should pass Travis (make sure to read the inline comment before).

deniak:
@astorije thanks for spotting the bug. Should be fixed now. Regarding your questions:
- I think the check should be done in echidna. The source is part of the token generation so imo, it should be checked at the same time
- we are indeed checking that the url provided starts with the source stored in the db. I had the same concerns about having something like `http://www.github.com` that will allow publication of anything under github but since the token registration is restricted to team contact for now, I think it's fine. We could try to force 1+ level in the source when requesting a token, but it may not work for specs not developed under github.

astorije:
It looks better indeed :)

- I don't understand what you mean by " The source is part of the token generation so imo, it should be checked at the same time", but my question was badly worded anyway. What I meant is that this check can be done in echidna, where you placed it, or on the Symfony side, in the request to the `global.TOKEN_ENDPOINT`. The result of the check would be available in the HTTP response. Is that clearer?
- Indeed, we cannot enforce the 1+ level. My question then could be: why don't we check that the source matches *exactly*?

deniak:
@astorije
- I understand now. You would pass a new parameter to the endpoint and symfony will compare to what's in the db. I suggest we leave it as is for now and create an issue for v1.5, but +1 on your idea.
- I don't think we can say the source should match exactly. People will use tags/branches etc...

tripu:
I created those two new issues for `v1.5` guys, and I think we can merge and close this one for now.
I'll let you do that @astorije, since you're assigned to it.

astorije:
Thanks @tripu and @deniak.
This implementation is indeed good enough for now. I'll merge.

astorije:
Rebased with master...

astorije:
... and now merged.

PR #33, opened 2015-01-23
Simulated token checker

plehegar:


astorije:
Action @astorije: Rebase and fixup last commit.

PR #35, opened 2015-01-26
Added "token" parameter to HTML form; plugged into "orchestrator".

tripu:


tripu:
This is the first half of the work in the 2nd row in [the table of pending tasks](https://www.w3.org/Team/wiki/Publications_workflow_design#Summary); *"publication request form"*.

tripu:
I personally favour having gutters as a conservative rule; not everyone has always the luxury of a wide monitor (or an external monitor at all).

We are indenting with 2 and 4 spaces, both in Specberus and here (sometimes 3, but those must be mistakes). 2 spaces seem to be more common here; 4 spaces are the rule in Specberus. But it's not really enforced (yet).

In any case, this underlines the need to take up again that definition of a common coding style. @TODO.

astorije:
Actually, only app.js doesn't have 2 spaces anymore on echidna.

tripu:
OK, then the ones I quickly found ([1](https://github.com/w3c/echidna/blob/master/test/lib/testserver.js#L17), [2](https://github.com/w3c/echidna/blob/master/test/lib/utils.js#L36)) are the odd mistakes. It's still 4 sometimes [in other languages in Echidna](https://github.com/w3c/echidna/blob/master/stubs/third-party-checker.sh#L8), though. Maybe we should unify that too once we reach consensus about coding style.

Interestingly, [Echidna was born with 4-space-indentation](https://github.com/w3c/echidna/commit/82390348596a138047d040124013da432d39abc3) (and with those ugly commas at the beginning of lines in multiple-var declarations) to be consistent with its predecessor Specberus&nbsp;&mdash;&nbsp;until, somehow, that changed&hellip; hum ;)

astorije:
Sorry, I think you meant that [Echidna was born with 2-space indentations](https://github.com/w3c/echidna/blob/82390348596a138047d040124013da432d39abc3/app.js). But I think fighting about 2-space indentation vs 4-space indentation was voted "Most Unproductive Fight in Computer Science" somewhere I just made up, so I won't fight :P

Regarding ugly-comma formatting, I plead guilty for getting rid of them. You're welcome :D

deniak:
My bad for totally re-indenting the files. Tripu and I started to switch the order of the jobs and I ended up re-indenting everything.
I guess we'll have to come up with coding style for v1.5. We throw so many things at v1.5, it's going to take forever to reach that milestones :)

PR #39, opened 2015-01-26
Fix a few issues with the installation to CVS

tripu:
* Added new config var `DEFAULT_BASE_TR_PATH`; it should hold the (preferably absolute) path to a local copy of the CVS repo, up to the `/TR/` bit (inclusive). Example: `/foo/bar/cvs-w3c/WWW/TR/`.
* Added function `buildFinalTRpath` to build the path for the new spec (using the aforementioned variable).
* Fixed a bug in the name of a function, `trInstall` vs. `trInstaller`.
* Also, improved a bit the layout of the HTML form, and the token field is *not* masked any more.

NB: while we're temporarily using `cp -R` to test this step, it's necessary to append `/*` to the `source`; otherwise the temporary directory with the timestamp will be copied (instead of its *contents*).


app.js
```diff
@@ -220,7 +230,8 @@ function orchestrate(spec, isManifest, token) {
                                         spec.jobs['publish'].status = 'ok';
 
                                         spec.jobs['tr-install'].status = 'pending';
-                                        return trInstall(tempLocation, finalLocation).then(function () {
+                                        finalTRpath = buildFinalTRpath(baseTRpath, report.metadata.get('thisVersion'));
+                                        return trInstaller(tempLocation, finalTRpath).then(function () {
```
astorije: Thanks for catching that!

app.js
```diff
@@ -177,6 +178,14 @@ History.prototype.toJSON = function () {
     return this.facts.reverse().toJSON();
 };
 
+function buildFinalTRpath(baseURL, thisVersionURL) {
+
+    var w3URL = /https?:\/\/(www\.)?w3\.org\/tr/i;
```
astorije: :-1:

Regular expressions often obfuscate the code and here is a good example. Reading this function, I cannot tell what is its purpose. When ultra-optimization is not in order, readability should be the key. Could you provide a regex-free alternative?

In terms of conservatism for the replacement, `This Version` URLs are never under `https`, always provided with `www`, and case is important (everything lower-case except for `TR`, always uppercase) so I am not sure you need to take extra care here.

astorije:
> NB: while we're temporarily using `cp -R` to test this step

Just to clear things up, this is *not* temporary. `cp` will and should always be used for development and test purposes. This is why the private installation script we use on the prod server should have the same interface than `cp`.

app.js
tripu: I was just trying to be as flexible as possible with our inputs. The registration form will accept something like `http://WWW.w3.oRg...` as the URL, right? So it should still work further along the process, I think. That's the good thing about regexes&hellip;

tripu:
@astorije: I updated the PR with a few simple fixes, after testing the CVS script with Denis today.

app.js
astorije: @tripu, the registration should not work with `https://www.w3.org/TR/...`, `http://WWW.w3.org/tr/...`, http://w3.org/TR/...`, etc.
The only accepted this version is `http://www.w3.org/TR/...`.

astorije:
@tripu, I'm confused. Would you mind adding a line of comment in or above the function? I don't see what it is supposed to to at first glance :/

tripu:
@astorije, that one-line function seems no longer necessary, so I've removed it. Hopefully it's clear now without any comments.
I've also rebased this branch, so that you can merge directly, if you think it's OK.

deniak:
@astorije the reason we have a regex is because Specberus test against a [regex](https://github.com/w3c/specberus/blob/master/lib/rules/headers/dl.js#L89) as well. I know the regex is not the same but it allows https for instance. Imo, it doesn't hurt to have a regex and it's not like it's a very complicated regex. That being said, if you have a strong opinion against the regex, @tripu and I are fine with you removing it and merge the MR. Afaik, that's the last thing blocking the MR

astorije:
Looks good, so I merge as is.

However, I vote for removing the flexibility in both Echidna **and** Specberus before v1: as the webmaster, I would never accept to publish documents with URLs like `https://www.w3.org/TR/...`, `http://WWW.w3.org/tr/...`, `http://w3.org/TR/...`. The `https` is special because we might switch TR to HTTPS someday, but other formats are just editors' mistakes and must be treated as such.

PR #40, opened 2015-01-27
Add simple logging, appending events and facts to a local text file

tripu:
* The *History* functions are now in their own module.
* There is a new config variable `DEFAULT_LOG_LOCATION` (suggested value: `.`).
* The log filename is `<DEFAULT_LOG_LOCATION>/echidna-<VERSION>.log`.
* Messages are appended to the log file with an automatic timestamp.

This is the second half of the work in the 2nd row in [the table of pending tasks](https://www.w3.org/Team/wiki/Publications_workflow_design#Summary); *"publication request form"*.

### DIFF
--- a/app.js
+++ b/app.js
@@ -11,12 +11,12 @@ var exec = require('child_process').exec;
 var path = require('path');
 var Request = require('request');
 var Promise = require('promise');
-var Stack = require('immutable').Stack;
 
 var DocumentDownloader = require("./functions.js").DocumentDownloader;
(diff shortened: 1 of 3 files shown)


astorije:
As discussed on IRC, this breaks [purity](https://en.wikipedia.org/wiki/Pure_function) as it adds side effects and prevents testing, for example.
Another issue is that it mixes up usages: History is human readable only, where logging should be its own task.

I plan to refactor this later.

tripu:
Fair enough concerns, @astorije &mdash; sure we'll fix that when we can refactor.

PR #44, opened 2015-01-29
W3C prefix: replace flexible regex with a more strict string

tripu:
See [@astorije's comment](https://github.com/w3c/echidna/pull/39#issuecomment-71855010).

### DIFF
--- a/app.js
+++ b/app.js
@@ -170,7 +170,7 @@ function orchestrate(spec, isManifest, token) {
     spec.jobs['tr-install'] = new Job();
     spec.jobs['update-tr-shortlink'] = new Job();
 
-    var W3C_PREFIX = /^https?:\/\/(www\.)?w3c?\.org/i;
+    var W3C_PREFIX = 'http://www.w3.org';
 
     var date = new Date().getTime();
     var tempLocation = (argTempLocation || global.DEFAULT_TEMP_LOCATION) + path.sep + date + path.sep;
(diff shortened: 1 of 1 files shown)


app.js
```diff
@@ -207,30 +207,41 @@ function orchestrate(spec, isManifest, token) {
...
-                                            spec.jobs['tr-install'].status = 'ok';
-
-                                            spec.jobs['update-tr-shortlink'].status = 'pending';
-                                            return updateTrShortlink(report.metadata.get('thisVersion')).then(function () {
-                                                spec.jobs['update-tr-shortlink'].status = 'ok';
-
-                                                var cmd = global.SENDMAIL + ' ' + global.MAILING_LIST + ' ' + report.metadata.get('thisVersion');
-                                                exec(cmd, function (err, stdout, stderr) {
-                                                  if (err) console.error(stderr);
+                                        if (0 === report.metadata.get('thisVersion').indexOf(W3C_PREFIX)) {
```
astorije: I'm not sure to understand what is the purpose of this `if` statement. If the document makes it that far, then the *This Version* URL passes the checkers, and is therefore a valid `http://www.w3.org/TR/[...]`.
Am I being mistaken here?

app.js
```diff
@@ -207,30 +207,41 @@ function orchestrate(spec, isManifest, token) {
...
-                                        });
+                                        }
+                                        else {
+                                            spec.jobs['publish'].status = 'failure';
+                                            spec.jobs['publish'].errors.push(errors);
+                                            spec.history = spec.history.add('The document could not be published: ' + errors.map(function (error) {
+                                                return error.message;
+                                            }));
+                                            return Promise.reject(new Error("Failed the publication system"));
+                                        }
```
astorije: One caveat here: if we are entering this `else` block (from line 237 to line 244), then `if(errors.length === 0)` (defined at line 206) still applies, so you will always get the following history message: `The document could not be published: `.

Not sure what you should display instead as I was wondering the purpose of the check in my comment above.

astorije:
@tripu, I feel pretty bad about throwing you inside this awful pyramid of doom...

Dude, I take an oath to clean this mess after v1. This has to stop.
On a serious note, it's a piece of code very difficult to maintain and to review, so it's easy to throw a bug inside.
Actually, ... #48 !

astorije:
@tripu, ping? (cf my comments inline)

tripu:
@astorije, I removed that check as you suggested. Please review, and merge if appropriate.

PR #52, opened 2015-02-02
send emails notifications using webreq

deniak:


astorije:
@deniak, the [Linux man page](http://linux.die.net/man/1/mail) says that `-a` is for attachements and mentions `-r` for the `From:` field. Am I looking at the wrong man page?

deniak:
@astorije we are using [bsd-mailx](http://man.he.net/man1/bsd-mailx)
```
     -a      Specify additional header fields on the command line such as "X-
             Loop: foo@bar" etc.  You have to use quotes if the string con-
             tains spaces.  This argument may be specified more than once, the
             headers will then be concatenated.
```

PR #56, opened 2015-02-02
Add a web-based testbed that automatically submits a batch of publications to Echidna and monitors results

tripu:
Change `http://localhost:3001/` to make it a better tool for automated testing of the system.

The testbed lists all specs and some metadata about them. A rudimentary *console* and a simple *inspector* allow the user to track progress of submitted specs. A *settings* sections at the top of the page makes it possible to easily test instances of Echidna deployed in different hosts/ports. Specs can be examined one by one; their contents, metadata, and errors thrown at each stage of the publication process.

In addition to that:
* Load specs and their metadata asynchronously from a new JSON file (prerrequisite to better handle those test cases, their metadata and publication results form the test page).
* Add CORS support, to be able to restrict usage of the Echidna API to an approved set of clients; new configuration variable `ALLOWED_CLIENTS` used for this.
* Better handle errors when starting the servers (both Echidna and test server); specifically, RTEs when the port is already in use.
* Use jQuery for DOM manipulation and AJAJ.
* Start using JS' *strict mode* in a few more files.
* Update test assertions to adapt to the contents of the new dynamic test page.
* Update the documentation accordingly.


app.js
```diff
@@ -1,3 +1,4 @@
+
```
astorije: ?!
Please remove, unnecessary...

tripu:
@astorije, I've removed my old-fashioned `EOF` comments because I understand they're not common at all, and my reasons to use them (as delimiters when I'm scrolling on an editor, `less`ing on the terminal, or reading code on paper) are personal preferences and not solid arguments :)

Regarding blank lines around the body of functions, blocks and `return` statements, and at the beginning and end of files&nbsp;&mdash;&nbsp;I'm not changing them, at least not until we decide on a common coding style. Blank lines don't cost a thing, and make the code more readable.

I'm addressing your other comments inline.

test/lib/testserver.js
```diff
@@ -50,7 +63,14 @@ TestServer.start = function () {
     init();
   }
   do {
-    server = app.listen(port);
+    server = app.listen(port)
+      .on('error', function(err) {
+        if ('EADDRINUSE' === err.code) {
+          // Just continue trying.
```
tripu: Yes, having two blocks here isn't necessary. I just wanted to make it explicit with the comment why we're capturing and handling every possible error, except `EADDRINUSE`. It would look a bit weird otherwise. Just legibility.

test/views/behaviour.js
```diff
@@ -0,0 +1,193 @@
+
+/**
+ * Behaviour of the Echidna testbed.
+ *
+ * @author Antonio <antonio@w3.org>
+ */
+
+'use strict';
```
tripu: Why? `'use strict';` is code, and so is close to the code; the documentation comment at the beginning of a file documents the entire file (including that statement), not only the function/class following it.

test/views/behaviour.js
```diff
@@ -0,0 +1,193 @@
...
+function dumpObject(obj) {
+
+    inspector = inspector || $('pre#inspector');
+
+    if (!obj) {
+        inspector.text('');
+    }
+    else {
+        inspector.text(JSON.stringify(obj, null, 2));
+    }
```
tripu: Again, our preference about this isn't decided yet. Douglas Crockford and others advocate always wrapping blocks in curly brackets, to avoid common mistakes, and I think it's a good idea.

(Apart from that, having the condition and the statement like you suggest, together in the same line, is much harder to read, don't you think?)

test/views/behaviour.js
```diff
@@ -0,0 +1,193 @@
...
+        if (id === allSpecs[i].id) {
+            result = allSpecs[i];
+        }
+
+        i++;
+    }
+
+    return result;
+
+}
```
tripu: I totally :+1: on this if it's supported by the platform.
I wouldn't introduce new complexity or dependencies just to have nicer statements like that one, though&hellip;

astorije:
For the `EOF`s, that was not sarcasm, I was genuinely curious :) Not, a bit of sarcasm: reading code **on paper**?! :D

Regarding the tons of blank lines, I disagree. JavaScript is already a very verbose language, we do not need additional unnecessary blanks. In particular, it makes scrolling back and forth almost mandatory when reading on screen (:P) and that defeats the purpose of trying to improve readability :-)

/me looks at the comments inline

test/lib/testserver.js
astorije: Fair concern, but it isn't incompatible:

```js
// We keep trying on errors only if because port already in use
if ('EADDRINUSE' !== err.code)
```
(or better wording)

Having an empty `if` block is really odd...

test/views/behaviour.js
astorije: Makes complete sense, I'll try to remember this convention when editing code.

By the way... please remove line 1, empty :D Don't tell me this improves readability :-)

test/views/behaviour.js
astorije: Douglas Crockford recommends that when there is more that one line, for reasons that the JS interpreter isn't always nice when you don't.

And no, I don't think it's harder to read. Actually, I think it's way easier to read when the `if` condition is short (aka when condition + execution < 80 chars). I prefer that billion times over scrolling all over the place when reading a file. Here, the function is **14 lines** when it could be half! Come on...

test/views/behaviour.js
astorije: Right, nice code is overrated. You know, the "What vs. How" argument is not here to show your code during aesthetics competitions, it helps to read, debug, to maintain, ...

So approved polyfill (with potential ugly code, nobody cares) + nice and readable function boy, that's the way to go!

tripu:
@astorije, I just changed the style of two of these blocks according to your comments.

(Maybe I wasn't clear before about the block and brackets thing. Crockford, among others, advises always using curly brackets, and always having the body of the block wrapped in brackets; [it's not a matter of how many statements, or about short statements following the condition in the same line](https://github.com/douglascrockford/ADsafe/commit/2cac52850aa6d45a4e4678134438a6af330cd3cf#diff-501682db26572ba7cb00fdbb34812204R1212). That's the approach I was supporting. I've changed it in this case to follow your style, nevertheless.)

Now, can we *please* have this merged, @plehegar? By the time it makes it into `master`, we'll have to rebase every other development branch&hellip;

PR #60, opened 2015-02-03
Generate unique id to identify the job

### LINKED ISSUE #55: Restrict “status” method to give info about particular publication jobs (instead of dumping all) and have the “request” method to return the ID of the new job
For privacy/security reasons, the HTTP GET method `status` of the API should *not* return all details about all publication jobs since the system was last started. Instead, it should accept an array of job IDs, and return info only about those jobs.

&hellip;which also means that the POST method `request` should always return the ID of the new publication job the user just submitted, instead of returning always a meaningless `200`&hellip;

deniak:
Fix #55

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ Then simply open [`http://localhost:3000`](http://localhost:3000) on your web br
 
 (with npm >= 2)
 ```
-$ npm start [-- STAGING_PATH [HTTP_BASE_URI [PORT]]]
+$ npm start [STAGING_PATH [HTTP_BASE_URI [PORT [RESULT_PATH]]]]
 ```
 
 Meaning of positional parameters:
@@ -52,6 +52,7 @@ Meaning of positional parameters:
 (Default `http://localhost/trstaging/`.)
 3. `PORT`: where Echidna will be listening for publication requests.
 (Default `3000`.)
+4. `RESULT_PATH`: local path where Echidna will dump the results of publication requests in JSON format.
 
 Examples:
 
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ Then simply open [`http://localhost:3000`](http://localhost:3000) on your web br
 
 (with npm >= 2)
 ```
-$ npm start [STAGING_PATH [HTTP_BASE_URI [PORT [RESULT_PATH]]]]
+$ npm start [-- STAGING_PATH [HTTP_BASE_URI [PORT [RESULT_PATH]]]]
 ```
 
 Meaning of positional parameters:
--- a/app.js
+++ b/app.js
@@ -268,15 +268,11 @@ function orchestrate(spec, isManifest, token) {
                                             }, function (err) {
                                                 spec.jobs['update-tr-shortlink'].status = 'error';
                                                 spec.jobs['update-tr-shortlink'].errors.push(err.toString());
-                                                spec.status = STATUS_ERROR;
-                                                dumpJobResult(resultLocation, spec);
                                                 return Promise.reject(err);
                                             });
                                         }, function (err) {
                                             spec.jobs['tr-install'].status = 'error';
                                             spec.jobs['tr-install'].errors.push(err.toString());
-                                            spec.status = STATUS_ERROR;
-                                            dumpJobResult(resultLocation, spec);
                                             return Promise.reject(err);
                                         });
                                     }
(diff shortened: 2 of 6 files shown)


views/index.html
```diff
@@ -163,7 +163,7 @@
              token: $('#token').prop('value'),
              decision: $('#decision').prop('value')},
             function(data) {
-              window.alert(data);
+              window.alert("Follow the status of the request using the token: " + data);
```
tripu: Better to talk about *ID* instead of *token* here, to avoid confusion?

tripu:
@deniak, note that this PR will not close #55 as it is now; only *half* of that issue&hellip;

tripu:
My changes over @deniak's initial PR:

* Improvements in `app.js`:
  * Error-prone string literals replaced with capitalised vars (pseudo-constants).
  * When the result is a simple text, or an ID, there is no need to wrap that in a JS object; now returning that instead.
  * Better HTTP status codes:
    * For `request`:
      * `202`: `accepted`.
      * `200`: `OK`, but that URL is already being processed.
      * `500`: missing parameters.
    * For `status`:
      * `200`: `OK`.
      * `400`: missing parameters.
      * `404`: job `not found`.
* Web UI updated to reflect these changes:
  * New signature of two methods.
  * Keeping in memory the IDs of all jobs submitted so far.
  * Invoking `status` as many times as jobs submitted so far.
  * Dealing with different HTTP status codes and response texts.
* Documentation (`README`) updated to reflect these changes; specifically, the new `RESULT_PATH` parameter.

@astorije, passing the baton to you now. I realise that the status codes in particular are messy and not coherent across both methods; feel free to fix them (just remember to update the web client accordingly :)

astorije:
@tripu, could you rebase your branch before I review the PR please? I have a couple of other things to do, so by the time I can read carefully your code, it will be rebased (therefore integrated with the current code and mergeable on Github). Thanks!

deniak:
@astorije code rebased. @tripu I changed the way we dump the error results. It's now handled directly in the `catch` instead of the `else`s.

astorije:
Thanks @deniak, I'll have a thorough look tomorrow.

tripu:
(Rebased again after the latest changes to `master`.)

tripu:
@astorije, FYI, @ylafon said today:

> POST -> 303 is also quite common http interaction
> usually done for factories
> like 'create new session'
> or... new publication ;)
> 303 redirects to something using GET, easier to monitor status

README.md
```diff
@@ -41,7 +41,7 @@ Then simply open [`http://localhost:3000`](http://localhost:3000) on your web br
 
 (with npm >= 2)
 ```
-$ npm start [-- STAGING_PATH [HTTP_BASE_URI [PORT]]]
+$ npm start [STAGING_PATH [HTTP_BASE_URI [PORT [RESULT_PATH]]]]
```
astorije: As far as I understood, the `--` is mandatory, [as explained here](http://stackoverflow.com/a/14404223/1935861). Did you experience the opposite?

app.js
```diff
@@ -74,30 +84,28 @@ app.post('/api/request', function(req, res) {
...
+            'history': new History(),
+            'status': STATUS_STARTED
+        };
+
+        orchestrate(requests[id], isManifest, token).then(function () {
+            console.log('Spec at ' + url + ' (decision: ' + decision + ') has FINISHED.');
+        }, function (err) {
+            console.log('Spec at ' + url + ' (decision: ' + decision + ') has FAILED.');
+        });
+        res.status(202).send(id);
```
astorije: Kudos for using a `202`! It clearly makes more sense than a `201` (but the publication system, our Symfony side, should still return a `201` when creating the resource in the DB - different topic).

Regarding `202` vs `303`, I'm not sold. You can change if you want, but I would suggest to leave it as is and switch after more feedback.

astorije:
Regarding the argument of the `send` methods, since we are building a JSON API, shouldn't we be returning JSON objects instead of strings? I am not suggesting at this stage, just wondering. Again, this can be adapted later if we need to.

astorije:
Hey, good job overall! :+1: 

I left a couple of comments inline for your consideration. Also, 3 super minor commits, feel free to include them or to cherry-pick them if you don't like.

deniak:
@astorije thanks.
- returning JSON instead of string. We discussed that with @tripu and ended up saying it's not a big deal for now and we'll see if people want JSON in the future. Easy enough to change
- HTTP code managed in our symfony bundles. Agree, we should take some time to review what's returned.

README.md
tripu: Yes.
`npm start ../specberus/staging/ http://localhost/specberus/staging/ 3030` works for me&hellip;
What am I missing?

app.js
WebAppsWG: 202 mean that you accepted the job, but don't give information about the status.
303 redirect to the status, in that case you will see the status (starting with the successful creation, of course).

ylafon: 202 mean that you accepted the job, but don't give information about the status.
303 redirect to the status, in that case you will see the status (starting with the successful creation, of course).

astorije: Hi @ylafon ! Could you give an example of how you see the output of such request? That'd be helpful.

PR #61, opened 2015-02-03
fix documentation using npm start

### LINKED ISSUE #57: npm start doesn't take args
README.md says one can run `npm start [STAGING_PATH]`, but npm doesn't seem to make it possible to pass arguments to the command.

(actually [npm2 might](https://github.com/npm/npm/pull/5518), but I don't think we should assume npm2 unless we must)

dontcallmedom:
close #57


astorije:
Thanks @dontcallmedom!

PR #62, opened 2015-02-03
support for interfacing with a separate spec-generator service

dontcallmedom:
For specs that needs a "build" step before they can be published (e.g. respec-based specs), we agreed back in November that having an optional parameter in the manifest to download the files through an external service would be useful.

This patch enables this.


plehegar:
This looks good to me. Only remaining part is to update respec on w3.org

PR #66, opened 2015-02-03
Extract publisher from app.js and make it testable and tested

astorije:
@deniak, @tripu, this cleans up some piece of code. Feel free to merge it now or to wait after v1 if you are not comfortable doing it now.

In a nutshell, the `Publisher` uses an abstracted `Service` (essentially an `object` that implements a `post` function) to publish. The `publish` function can be easily tested thanks to that.
Of course, the implemented JsonHttpService is not tested but that's not possible with unit testing. 

Regarding tests, I am not testing the way the metadata are treated, but what and how to efficiently test here... It will deserve its own PR, but I'd be happy to have feedback. @plehegar, Chief Engineer in Test, any idea?

I also used a [JSDoc](http://usejsdoc.org/) syntax for this. I am planning to use it more and more in the future. I am not forcing a syntax obviously, but it seems better than nothing :)


astorije:
Actually, I forgot to replace something. You can still review what has done so far and then just the new commit afterwards, if you want.

tripu:
On a quick pass this looks good, @astorije! I'll review more carefully later; no problem for me merging while still in `v1`.

:+1: about adopting JSDoc (or YUIDoc); we should be using them already.

astorije:
Ahoy, thanks @tripu! As said, I'll commit one last change before it can be merged. But don't see that as blocking v1: if I don't finish on time, I don't finish on time, no worries.

Regarding the doc, I'm not pushing anyone to do it at this stage. I just figured I could try it and see how useful and readable it is for the rest of the group. I don't mind the doc system to use either. I chose JSDoc because it seems to be by far the most used but it has a lot of drawbacks. One good thing is that its syntax seems to be compatible with most systems so even if we choose something else that won't be a big deal.

astorije:
Here you go, it can be reviewed and merged whenever you guys want. Sorry for re-pushing the whole history, I had to rebase with `master`.

plehegar:
looks ok to me

astorije:
\o/
Thanks @tripu and @plehegar!

PR #51, opened 2015-02-02
Apply fixed major/flexible minor+patch for dependencies and devDependencies + dependency cleanup

astorije:
This commit also applies the version strategy suggestion **(1b)** explained in #50.

PR #71, opened 2015-02-05
Restore Grunt development dependencies to be able to manage environments and do builds

tripu:
Restore all Grunt-related development dependencies that we got before they were removed in 28aa84c140121c3eb037eb2f6edc42dd1cbf9651. For the moment, just to be able to run `grunt build` and switch easily between production and development environments.

(There are probably a few unneeded Grunt dependencies; we'll have to prune that soon.)


deniak:
@tripu, we should use N.x.x for versions

astorije:
No! Please **don't**!

And BTW, whenever a version has a `0` major, you should use `0.n.x` as per `man semver`:

> Many authors treat a 0.x version as if the x were the major "breaking-change" indicator.

tripu:
What do you mean, @astorije? No to using Grunt?

About specifying up to *minor* for those modules: OK; then I'm reverting my previously reverted change, and using `0.<whatever>.x`.

astorije:
Well, when you spend time on cleaning the dependencies and [that gets approved](https://github.com/w3c/echidna/pull/51), it's kind of frustrating to see it reverted and merged overnight!

I'm fine using grunt if there is a real value to it. That is, if it is explained here or somewhere else and documented on the README or the wiki. So, no to blindly adding *20 dependencies*. We will probably not use most of them. Regarding SASS, SVG-min, CSS-min, asset management being over-engineered, ... no! This is a small project, we don't need the extra overhead of automated CSS optimization to win half a kilobyte (assuming everything is gzipped).

I'm not saying we shouldn't provide that in the future, but at the moment it's clumsy, over-engineered and not documented. In a nutshell, unnecessary.

PR #77, opened 2015-02-05
Fix #64...

### LINKED ISSUE #64: Crash when server is not reachable
If the URL of the target document is not reachable, echidna crashes instead of indicating failure to retrieve.

astorije:
... More or less. #64 was actually fixed by 6b7d231 but it was not its purpose. This PR makes it clearer in terms of message and provides a couple of additional tests.

### DIFF
--- a/lib/document-downloader.js
+++ b/lib/document-downloader.js
@@ -10,11 +10,15 @@ var DocumentDownloader = function() {};
 
 DocumentDownloader.fetch = function (url) {
   return new Promise(function (resolve, reject) {
-    Request.get(url, function (error, response, body) {
-      if (error) reject(error);
+    Request.get(url, { timeout: 30000 }, function (error, response, body) {
+      if (error) {
+        reject(new Error(
+          'Fetching ' + url + ' triggered a network error: ' + error.message
+        ));
+      }
       else if (response.statusCode !== 200) {
         reject(new Error(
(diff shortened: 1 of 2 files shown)


dontcallmedom:
can you add a test for the network error case? e.g.
```javascript
    it('should reject if the server is not reachable', function () {
      var notReachable = DocumentDownloader.fetch('http://youdbetternotexit/');
      return expect(notReachable).to.eventually.be.rejectedWith(/network error/);
    });
```

tripu:
@dontcallmedom, I added the test you suggested.

deniak:
@astorije @dontcallmedom I think that MR is now complete. I'll merge it so we have the fix deployed. Let me know if I missed something.

astorije:
@tripu, I was on it...

astorije:
@tripu, [I added a test for fetchAndInstall](https://github.com/w3c/echidna/commit/2f9734c9245aa319d2035366b8eb5ebda4006c4d#diff-c1129c8b045390789fa8ff62f2c6b4a9R180), for completeness. I merged it directly with @deniak's approval since this was already merged. No more commits planned :)

tripu:
I'm sorry about my interference with this one, @astorije :( Timezone differences and the need to meet our deadlines made us step in each others' shoes a few times&hellip;

PR #80, opened 2015-02-06
Added automatic detection of manifest and removed isManifest parameter

plehegar:
Instead of a parameter, let's do:

if (content.trim().charAt(0) !== '<') {
  // it's a manifest
}

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -144,11 +144,11 @@ describe('DocumentDownloader', function () {
     });
   });
 
-  describe('fetchAndInstall(url, dest, isManifest)', function () {
+  describe('fetchAndInstall(url, dest)', function () {
     var promise;
 
     before(function() {
-      promise = DocumentDownloader.fetchAndInstall(server.location(), '/tmp/testechidna', false);
+      promise = DocumentDownloader.fetchAndInstall(server.location(), '/tmp/testechidna');
     });
 
     after(function(){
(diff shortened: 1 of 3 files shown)


plehegar:
It would be nice to merge this one by Monday and update the documentation

test/test.js
```diff
@@ -197,8 +197,7 @@ describe('DocumentDownloader', function () {
     it('should read a manifest and install its content', function () {
       return DocumentDownloader.fetchAndInstall(
         server.getMetadata('navigation-timing-2').location + 'W3CTRMANIFEST',
-        '/tmp/testechidnaManifest',
-        true
```
astorije: @plehegar, could you remove the `false`s line 183 and 192 as well please?

astorije:
@plehegar, could you update the HTML interface as well please?

The overall idea is fine by me. By making the detection automatic instead of part of the API, we can change or improve the detection algorithm later without affecting the API.

@deniak, @tripu, I will let you merge if you agree with the feature. We can discuss about it if you don't but I won't merge without you :)

tripu:
As I told @deniak, I'm fine with merging this for the time being. But ideally, I would like to rethink the method `request` and its signature to make it more robust, and easier to extend.

As we start supporting more preprocessors (each with its own parameters) and adding options to the tool, we'll need a better way to pass parameters to the API. I'm thinking a single `string` parameter, and storing there JSON with all configuration needed. Will discuss with you guys in a separate issue, or on the wiki&hellip;

deniak:
@plehegar @astorije @tripu it's merged and deployed. The documentation has also been updated.
Regarding the web interface, it's handled in a separate [branch](https://github.com/w3c/echidna/tree/web-interface) and I removed the checkbox.

For the future, I agree with @tripu. We should find a way to handle all kind of submission more easily and make the system more flexible.

astorije:
Hey @deniak and @tripu! I essentially agree with you, but I think now is too premature to decide something like this. I believe we need production experience and feedback to better handle these cases.

The point of this PR is to make the v1 API simpler. The way we handle the manifest is now handled under the hood, which is good. We can knit a nicer hood later :-)

Issue #6, opened 2014-12-23
Add a property file to replace hardcoded values


deniak:

astorije:
Some [follow-up] coming from @darobin.

tripu:
Done. We already have [`config.js.example`](https://github.com/w3c/echidna/commits/master/config.js.example), which users should duplicate as `config.js`, and then edit.

--

Related: maybe getting rid of CLI arguments altogether (@astorije) or improving them, eg using [`nopt`](https://www.npmjs.com/package/nopt) (@darobin).

astorije:
> Related: maybe getting rid of CLI arguments altogether (@astorije) or improving them, eg using [`nopt`](https://www.npmjs.com/package/nopt) (@darobin).

Yep, that's why I didn't close the issue right away :-)

tripu:
I think it's best to open separate issues in cases like this, @astorije. Otherwise we have to change the title and description of existing issues, and we agreed that is a bad idea. No?

astorije:
Done @tripu: #157

PR #84, opened 2015-02-15
Modular orchestrator

astorije:
:warning: **READY TO BE REVIEWED BUT NO TESTS YET** :warning:

One of the ideas here is that `app.js` should get as little intelligence as possible.

The result is an `Orchestrator` that controls the process of publication. This class manipulates a frozen (immutable) state, `RequestState`. Basically, everything the `Orchestrator` does is manipulate a state that is given as an argument of the computation so that the outer state is never modified as a side effect. That way, we avoid unexpected changes.

To help with this and abstracting, the code makes use of higher-order functions (functions given as arguments or returned by other functions). This separates concerns because the code for a function makes little assumption on what an argument is supposed to do in addition to its type (for example, if as an argument I give a function `f: Int => Promise.<Int>`, I don't really care what `f` is doing, I only care about what I feed it with and what I get in return). I'll keep improving the code to that extent to favor maintainability.

In the end, the main entry point is a recursive function that recurses on a function producing a promise until a predicate is reached. This will probably be hidden under the hood in a near future, but at least in the present state it allows us to do *something* (whatever it is, since it's an higher-order function) every time the request state is changed (that is, every time a step in the process makes it change). At the time of writing this, I am just displaying the new state on the console, but it is also a good place to consider persistence support.

### DIFF
--- a/app.js
+++ b/app.js
@@ -358,9 +358,22 @@ function orchestrate(spec, token) {
     });
   }
 
-  function runShortlink(specberusReport) {
+  function runShortlink(specberusReport, spec) {
+    spec.jobs['update-tr-shortlink'].status = 'pending';
+
     return specberusReport.then(function (report) {
-      return updateTrShortlink(report.metadata.get('thisVersion'));
+      return updateTrShortlink(report.metadata.get('thisVersion'))
+        .then(function () {
+          spec.jobs['update-tr-shortlink'].status = 'ok';
+
+          return Promise.resolve();
+        }).catch(function (error) {
+          spec.jobs['update-tr-shortlink'].status = 'error';
+          spec.jobs['update-tr-shortlink'].errors.push(error.toString());
+          spec.history = spec.history.add('An error occurred while updating the shortlink.');
+
+          return Promise.reject(error);
+        });
     });
   }
 
@@ -400,7 +413,7 @@ function orchestrate(spec, token) {
       return runTrInstaller(specberusReport, tempLocation, spec);
     })
     .then(function () {
-      return runShortlink(specberusReport);
+      return runShortlink(specberusReport, spec);
     })
     .then(function () {
       return finishTasks(specberusReport, spec, resultLocation);
@@ -424,15 +437,8 @@ function orchestrate(spec, token) {
 
     spec.jobs['retrieve-resources'].status = 'ok';
     spec.history = spec.history.add('The file has been retrieved.');
-
-              spec.jobs['update-tr-shortlink'].status = 'pending';
-
-                spec.jobs['update-tr-shortlink'].status = 'ok';
                 spec.history = spec.history.add('The document has been published at <a href="' + report.metadata.get('thisVersion') + '">' + report.metadata.get('thisVersion') + '</a>.');
                 spec.status = STATUS_SUCCESS;
-
-                spec.jobs['update-tr-shortlink'].status = 'error';
-                spec.jobs['update-tr-shortlink'].errors.push(err.toString());
 
     spec.history = spec.history.add('The document could not be retrieved.');
     spec.jobs['retrieve-resources'].status = 'error';
--- a/app.js
+++ b/app.js
@@ -12,7 +12,7 @@ var Uuid = require('node-uuid');
 
 var History = require('./lib/history');
 var Orchestrator = require('./lib/orchestrator');
-var SpecberusWrapper = require('../functions.js').SpecberusWrapper;
+var SpecberusWrapper = require('./functions.js').SpecberusWrapper;
 
 // Configuration file
 require('./config.js');
--- a/app.js
+++ b/app.js
@@ -2,6 +2,7 @@
 
 console.log('Launching…');
 
+var exec = require('child_process').exec;
 var meta = require('./package.json');
 var express = require('express');
 var compression = require('compression');
@@ -23,6 +24,8 @@ var STATUS_STARTED = 'started';
 var app = express();
 var requests = {};
 var port = process.argv[4] || global.DEFAULT_PORT;
+var argTempLocation = process.argv[2] || global.DEFAULT_TEMP_LOCATION;
+var argHttpLocation  = process.argv[3] || global.DEFAULT_HTTP_LOCATION;
 var argResultLocation = process.argv[5] || global.DEFAULT_RESULT_LOCATION;
 
 app.use(compression());
@@ -71,6 +74,21 @@ app.get('/api/status', function (req, res) {
   else res.status(400).send('Missing required parameter “ID”.');
 });
 
+function Job() {
+  if (typeof this !== 'object') {
+    throw new TypeError('Job must be constructed via new');
+  }
+
+  this.status = '';
+  this.errors = [];
+}
+
+function dumpJobResult(dest, result) {
+  Fs.writeFile(dest, JSON.stringify(result, null, 2) + '\n', function (err) {
+    if (err) return console.error(err);
+  });
+}
+
 app.post('/api/request', function (req, res) {
   var url = req.body ? req.body.url : null;
   var decision = req.body ? req.body.decision : null;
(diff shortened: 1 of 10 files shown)


tripu:
Good luck with this one, @astorije&hellip; ;) Feel free to assign it to me when you're ready.

astorije:
Cancelling Travis CI as style checker won't pass...

astorije:
Pfiou, I just sucessfully rebased my 41 commits, that was something :-)

@tripu, I think I am more or less done here!
Before giving the go, there is one thing I'd like to confirm with you: this refactoring will help dealing with storing things. While developing, I was just `console.log`ging what I was getting as an output, but I figured I might have lost some of your persistence-related code. After checking though, the original `app.js` just printed something in the console. In my understanding, the only thing that appears on the storage layer is appending to a log in the `History`. Could you confirm this? We can have a quick chat if this doesn't make sense.

If the answer to the previous question is no, then you can start reviewing. Meanwhile, I will provide tests now that this architecture offers it \o/

I have put a lot of love in this, I hope you'll like it :-)

tripu:
@astorije, this LGTM!

I see this build failed, but only because code coverage dropped a little bit. No big deal now.

I think you are right in that the only &ldquo;output&rdquo; here was done through the module `History`. So if I understood the logic in your comment above, that means we're OK to continue (read: merge), right?

I've tried to follow your commits and the diff, but yeah, it's difficult :¬/ For what I can see, the core logic and the flow of steps and checks is preserved (and obviously tests pass), so I won't try to dig very deep.

About tests: I'd say we merge this right now, as is; and work on tests in a separate PR. I fear that if you, @plehegar or I tackle tests for this in this same branch and without merging to master first, the snowball is going to keep of growing. I predict new rebases and an even messier diff to review&hellip; Maybe I didn't understand you, and that is what you were suggesting anyway?

Please merge yourself (or grab @deniak if you want a third opinion; but I say go).

Good job :¬)

astorije:
> I see this build failed, but only because code coverage dropped a little bit. No big deal now.

98% is still a decent number :P
And the funny part is that when I'll start adding tests, the coverage will decrease as the new files in `lib/` are [not parsed yet](https://coveralls.io/builds/2338374) (click on *All*).

> I think you are right in that the only &ldquo;output&rdquo; here was done through the module `History`.

Okay, I have some ideas to improve this then :-)

> So if I understood the logic in your comment above, that means we're OK to continue (read: merge), right?

Yes.

> I've tried to follow your commits and the diff, but yeah, it's difficult :¬/ For what I can see, the core logic and the flow of steps and checks is preserved (and obviously tests pass), so I won't try to dig very deep.

Indeed, I essentially just made things modular and scalable. My worry was that I might have lost some existing behavior, mostly persistence and side-effects (since most of what's added in `lib/` is side-effect-free).

> About tests: I'd say we merge this right now, as is; and work on tests in a separate PR. I fear that if you, @plehegar or I tackle tests for this in this same branch and without merging to master first, the snowball is going to keep of growing. I predict new rebases and an even messier diff to review&hellip; Maybe I didn't understand you, and that is what you were suggesting anyway?

Well, I would never suggest merging without tests or biking without a helmet :D
But I agree that this PR is too big to be postponed further. A good outcome is that we should be able to work on the main loop in small bits from now on.
So okay to merge without tests, but if you're okay with that, let's leave it a couple of weeks before the next release to give some time to tests.

> Please merge yourself (or grab @deniak if you want a third opinion; but I say go).

Here is my proposal: I'll merge only tomorrow. If @deniak, @darobin, @tabatkins (thanks for your input!) or @dontcallmedom has a free eye to spare on this (partially, even) by then, that's cool. Otherwise, no worries.
Meanwhile, I'll see if I can add a test or too and/or see if something needs fixing. I'll merge afterwards.

> Good job :¬)

Thanks :-)

app.js
```diff
@@ -2,37 +2,32 @@
...
+var RequestState = require('./lib/request-state');
 var SpecberusWrapper = require('./functions.js').SpecberusWrapper;
-var ThirdPartyResourcesChecker = require('./lib/third-party-resources-checker');
-var TokenChecker = require('./functions.js').TokenChecker;
 
 // Configuration file
 require('./config.js');
 
 // Pseudo-constants:
 var STATUS_STARTED = 'started';
```
dontcallmedom: this is brittle (strings redefined across modules); I would make it a "static" property of RequestState. (i.e. add RequestState.started = "started", etc in lib/request-state.js)

app.js
```diff
@@ -2,37 +2,32 @@
...
-var STATUS_ERROR = 'error';
-var STATUS_SUCCESS = 'success';
 
 var app = express();
 var requests = {};
+var port = process.argv[4] || global.DEFAULT_PORT;
```
dontcallmedom: the changes are many enough that semantic-free changes should be avoided :)

lib/orchestrator.js
```diff
@@ -0,0 +1,354 @@
...
+Orchestrator.prototype.runDocumentDownloader = function (url, tempLocation) {
+  return new Map({
+    name: 'retrieve-resources',
+    promise: DocumentDownloader.fetchAndInstall(url, tempLocation)
+      .then(function () {
+        return new Map({
+          status: 'ok',
+          history: 'The file has been retrieved.'
+        });
+      }).catch(function (error) {
```
dontcallmedom: all the subsequent run* function  have a very similar catch handler; probably some room for DRY

lib/orchestrator.js
```diff
@@ -0,0 +1,354 @@
...
+
+/**
+ * Run the next step of the orchestration, for a given state.
+ *
+ * @param {RequestState} state - A state for a specific publication request
+ * @returns {List.<Promise.<RequestState>>} A list of promised states that are fulfilled at different stages of each step
+ */
+Orchestrator.prototype.next = function (state) {
+  if (state.hasJobStarted('retrieve-resources')) {
+    return this.runStep(
```
dontcallmedom: instead of repeating this.runStep(…)(state), put the step to be run in a variable, and call runStep() only once?

lib/request-state.js
```diff
@@ -0,0 +1,127 @@
...
+ * @param {Map.<string, Job>} jobs - A map of all the jobs related to this request
+ * @param {History} history - A human-readable history for this request
+ * @param {Map.<string, *>} metadata - The metadata extracted by Specberus for this request
+ */
+function RequestState(status, jobs, history, metadata) {
+  if (typeof this !== 'object') {
+    throw new TypeError('RequestState must be constructed via new');
+  }
+
+  this.status = typeof status !== 'undefined' ? status : '';
```
dontcallmedom: why typeof x!== 'undefined' rather than x !== undefined (sorry if this is obvious)?

app.js
astorije: Yep, I am planning to consolidate this very soon. Thanks!

app.js
astorije: I'm sorry... What?! :-)

lib/orchestrator.js
astorije: Yes! I am also planning to abstract the `run*` functions soon (but most likely post-merge) :-)

app.js
dontcallmedom: you just moved this from line 35 to 28, right?

lib/orchestrator.js
dontcallmedom: to be very clear, I strongly +1 merging first :)

lib/request-state.js
astorije: `typeof x!== 'undefined'` is one of these things that are generally recommended but that I have to look up to remember why.

In that case, this is a bit of a stretch, but `undefined` can be overwritten (apparently, this is not true since ES5 though) while `typeof` is a built-in operator that always returns a `string`.

[This page](http://stackoverflow.com/questions/3390396/how-to-check-for-undefined-in-javascript) has some interesting stuff. I will not argue over what we should use: I am simply not experienced enough to take a stand. Please let me know what you prefer and I'll align with it.

lib/orchestrator.js
astorije: Yes, I have some plans towards that direction indeed. The `next` function at the moment is gross anyway.
I'll start by adding your suggestion.

app.js
astorije: Actually, it bounced from file to file over the commits before coming back here, I might not have put it back to its original location indeed, sorry :-)

I'll make a pass overall to get rid of these, thanks for reminding!

lib/request-state.js
darobin: The `typeof` thing is only required for global vars (to avoid reference errors), so it's not needed here indeed. That said, it's not uncommon to see it done that way because double-safe habits die hard (I know it happens to me a fair bit).

lib/orchestrator.js
astorije: Good to know your position on this, thanks!

app.js
darobin: More generally (but this is completely for another PR), once you start having `process.argv[5]` I would suggest starting to use a CLI options parser (typically `nopt`). It can get painful to remember these things :)

lib/request-state.js
darobin: I forgot to add: if someone overwrites `undefined` on you, you actually want it to create a bug. Because that will give you a reason to then go break their fingers off with a sledgehammer, as they so totally deserve.

In general for these things I follow the simple rules in those two sections: http://contribute.jquery.org/style-guide/js/#equality

darobin:
I haven't looked very closely, but it looks like good stuff and I'm all for fast-merging of refactors that pass their tests. Otherwise things get really painful.

astorije:
Huge thanks to @dontcallmedom for reviewing this. Everything you highlighted is very valid and I'll make sure to go through all of them, whether it's prior to or after merging.

@tripu, most comments will be addressed after merging I think, but there are a couple of things I'd like to change before merging, so definitely let me merge if you don't mind (or wait to merge yourself) :-)

lib/request-state.js
astorije: Thanks @darobin for your input! So you simply recommend @dontcallmedom's version, right? I'm fine either way.

app.js
astorije: :+1: 
However, I would only commit to a config file rather than supporting both command line arguments *and* a config file...

astorije:
@darobin thanks again for the input. Although I can entirely break Echidna *and* pass the tests with my changes, you know :-)

I'll process comments of both of you tomorrow, fixing right away what I can fix and opening issues/preparing PRs for the rest. I am very glad overall that you all like it! Thanks :heart:

lib/request-state.js
darobin: Yup, I'd do what @dontcallmedom says. (It's usually a good idea :)

astorije:
Actually, after re-reading all comments, I am going to merge this now, open issues to document what was said and send PRs for these (sooner or later depending on the issue).

app.js
astorije: Et voilà, I forgot to change that before merging... Tant pis!

app.js
astorije: See #148

app.js
astorije: See https://github.com/w3c/echidna/issues/6#issuecomment-94041185

lib/orchestrator.js
astorije: See #149

lib/orchestrator.js
astorije: See #151

lib/request-state.js
astorije: See #150

tripu:
OK to remove the branch?

PR #90, opened 2015-02-20
Removed the dependencies list

marcoscaceres:
I'm not sure why they are listed - it's just adds clutter for people trying to get started. The are added automatically by npm, so who cares? I can see what the dependencies are when the thing installs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,11 @@
 
 Echidna is the central piece of software taking care of the new publication workflow at [W3C](http://www.w3.org/). The plan is for Echidna and related sub-projects (see below) to automatise the publication of new specs under [`http://www.w3.org/TR/`](http://www.w3.org/TR/).
 
+## OS-level dependencies
+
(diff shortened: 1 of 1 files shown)


tripu:
I think you're right about the npm dependencies, @marcoscaceres. Besides, we're changing them often and not doing a good job at keeping the `README` up to date&hellip;

Could you leave the *OS-level dependencies* section, though? Those are not installed automatically by npm, and I think those are useful to know for anyone planning to install the system.

astorije:
Hey @marcoscaceres, this is something I meant to do for ages, thanks for bringing that up again! I wanted to fulfill @tripu's requirements as well as some other things I wanted to change on the README, but your commit was outdated so I could not base #96 on your commit.
Therefore, I think this is addressed by #96 and can be closed.

marcoscaceres:
Closing, as it's not worth doing a rebase just for this. Will send separate PR.

PR #93, opened 2015-02-20
Note in README local setup of the full system is not possible

anssiko:
For reference, see https://lists.w3.org/Archives/Public/spec-prod/2015JanMar/0011.html


astorije:
@tripu, I saw you assigned yourself on this ticket, but I wanted to add this commit to the parent of #96's commit, which is why I merged it... Hope you don't mind :-)

PR #81, opened 2015-02-07
Add a CONTRIBUTING.md file to the project

astorije:
This creates a banner every time someone opens an issue or a pull request and gives explanations to contributors.

Along the way I made 2 guides that are pointed by this contribution reference ([Coding style guide](https://github.com/w3c/echidna/wiki/Coding-style-guide) and [Coding practices guide](https://github.com/w3c/echidna/wiki/Coding-practices-guide). I am in the process of finalizing them. I'll be happy to discuss these 2 but I figured it would make everyone save time by starting something and would also guide contributors rather than letting them in the blur. I'll be cleaning things only when everyone is okay with these and try to provide a linter configuration file.

I used *plenty* of references to make something as good as possible.
I gave a lot of love to these documents and I am only willing to improve them :-)
Hurray to free and open source projects! \o/

(I am not assigning anyone here as I'd love feedback from both @deniak and @tripu, but also from @plehegar on the "public-facing" side. I hope this won't create too much of a debate and can be merged on Monday at the same time than when we open the system to the public.)

PR #101, opened 2015-03-07
Quality assurance PR!

astorije:
- As discussed on #81, applies the style consistently across all JavaScript files in the project (except `Gruntfile.js`, minified files, config file and packages). There were some syntax bugs I discovered along the way that I fixed. Pfiou, that was a lot of changes...
- Add a bunch of [JSCS](http://jscs.info/) rules to make sure a consistent style is kept over time. I went through [aaaall the rules](http://jscs.info/rules.html) twice to make sure nothing gets forgotten!
- Make the style checker run before the tests to never forget to check it before committing, and to have Travis validating it.
- Choices made in `app.js` to apply the rules will probably seem odd, but I didn't bother making it too beautiful as work on #84 is pending.
- Add rules for [JSHint](http://jshint.com/) and fix the errors it reported
- Make the linter run before the tests for the same reasons as the style checker


astorije:
@deniak, assigning to you as you were the one suggesting cleaning up the sources should come with suggesting one.
I will rebase my other pending PRs as soon as this is merged.

deniak:
@astorije @tripu I'm assuming you guys discussed and agreed on the coding style conventions. I would prefer to have `"requireMultipleVarDecl": true` in `.jscs.json` though.
Anyway, I'll approve the PR so it won't block anything else.

Now that the coding style convention is in place, we should probably enforce the rules in a git pre-commit hook. What do you guys think?

astorije:
@deniak, thanks for merging this! Actually, @tripu and I haven't discussed about that since #81, if I am correct. And that's the thing actually: it's the kind of things you want to discuss about, but you'd rather spend time on things that are actually productive!
Anyway, please don't see that as stabbing you in the back, but as a way of efficiently and consistently enforce the style and practices. If I have done things right, all rules should be either mandatory, through `require`, or forbidden, through `disallow` (apart from the `YodaConditions` where I really have no idea what's best and if we should really require or disallow). That way, whenever the style evolves, it will be easy to detect non-compliant statements as they will break the checker by design, and they will also be easier to fix across the project.

Regarding `requireMultipleVarDecl`, actually, I personally hate that style. I don't see what is wrong with having multiple `var` statements because you don't save space anyway. Moreover, and that's what I have seen multiple times on other projects, it can create "quiet" bugs like this:

```js
var myVar1,
    myVar2;
    myVar3;
```

Et voilà, `myVar3` is now a leaky variable. Granted that the use of `'use strict';` *should* prevent that (unless it was forgotten for that file) and JSHint *should* be unhappy too, but that's still a useless risk when you can be explicit.
This also makes the git diff awkward when you change one of the variables, and then people try to "fix" this with ridiculous practices such as starting lines with commas...
Anyway, if you guys both want this to change, sure, but I don't see a good reason, and I prefer simplicity :-)

Regarding the pre-commit hook, as the [documentation says](http://www.git-scm.com/book/en/v2/Customizing-Git-Git-Hooks#Client-Side-Hooks), client-side hooks are not copied when cloning the repository. So it means that we need to add a section in the CONTRIBUTING file asking people to create the hook. I don't think people will bother creating a hook if they don't bother running tests before committing, which is required in the CONTRIBUTING file.

I will rebase all the PRs ASAP to have them merged quickly.

astorije:
Oh, I was forgetting an argument regarding the hook: sometimes you actually want to commit something that doesn't pass the tests and I think it can happen with the style or linter too. For tests, it might be that you submit code that doesn't pass tests yet because of a pending feature in a dependency. For the style checker or linter, maybe someday a contributor will want to help and will not know how to make them pass, but would still like to share so that we can help debugging...

PR #116, opened 2015-03-16
Add changelogs for both Echidna and Specberus, as of today

tripu:


tripu:
On hold.

@deniak suggested that it might be better to use a text file, or a markdown file, for the changelog(s). Also, he wonders &ldquo;if Echidna is a good place to host the changelog for Specberus&rdquo;.

I think this: Echidna is the core of the system, and its front page (and future web UI) serve as the face of the whole system. Thus the convenience of maintaining a friendly changelog there, at the &ldquo;base URL&rdquo; that users would likely check. Also, it is unlikely that Specberus will evolve separately (releases will be pretty much related and in sync) so summing up changes in Specberus *in the front page of Echidna* doesn't seem to me like an ugly dependency or anything&hellip;

Still, I'm open to better ideas.

astorije:
I agree with @deniak: changelogs for Specberus should be stored within the Specberus repository.

I actually opened a ticket on Specberus last week: https://github.com/w3c/specberus/issues/165. If we manage to generate the changelogs, this will be solved inherently, and we'll be able to redirect users to the `HISTORY.md` files, which is more than enough. Almost all open source projects who maintain a changelogs store it either as a CHANGELOG/HISTORY.md file in the repository or a tag notes on GitHub (I prefer the first solution but see the convenience of the second one).
That said, I haven't found a good way to generate this file yet, although I guess at the very least, a command involving `git log` could do the trick.

Hope that helps!

tripu:
/me moving the discussion to w3c/specberus#165 the, where it seems more appropriate.

PR #122, opened 2015-03-16
fix downloader sub-directories missing

### LINKED ISSUE #119: The document downloader fails when the manifest lists sub-directories
If the manifest contains files in sub-directories, the document downloader [doesn't create](https://github.com/w3c/echidna/blob/master/lib/document-downloader.js#L67) the sub-directory before downloading the resource. It ends up failing with the following error:
```
...
"jobs": {
    "retrieve-resources": 
{
    "status": "error",
    "errors": 
    [
        "Error: ENOENT, open '/u/echidna-staging/47aadb10-02fc-474a-9f1c-69b59f552b29//img/fingerprint.png'"
    ]
},
...
```

deniak:
Fix #119

### DIFF
--- a/lib/document-downloader.js
+++ b/lib/document-downloader.js
@@ -44,8 +44,8 @@ DocumentDownloader.installAll = function (destsContents) {
       if (!exists) {
         Mkdirp.sync(path);
       }
+      return DocumentDownloader.install(e[0], e[1]);
     });
-    return DocumentDownloader.install(e[0], e[1]);
   }));
 };
 
--- a/lib/document-downloader.js
+++ b/lib/document-downloader.js
@@ -5,6 +5,7 @@ var Request = require('request');
 var List = require('immutable').List;
 var Promise = require('promise');
 var Url = require('url');
+var Mkdirp = require('mkdirp');
 
 var DocumentDownloader = function () {};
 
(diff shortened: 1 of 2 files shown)


lib/document-downloader.js
```diff
@@ -38,6 +39,8 @@ DocumentDownloader.install = function (dest, content) {
 DocumentDownloader.installAll = function (destsContents) {
   // `e` is a tuple of [dest, content]
   return Promise.all(destsContents.toArray().map(function (e) {
+    var path = require('path').dirname(e[0]);
+    Mkdirp.sync(path);
```
astorije: This sync version is fine by me.
I don't think it takes a lot longer to execute but I'd be curious to see the difference. Meanwhile it makes the code easier to read than a callback-based async version. When (if!) `mkdirp` [supports Promises](https://github.com/substack/node-mkdirp/issues/86), we can probably do something like:

```js
return Mkdirp(path).then(function ()
  return DocumentDownloader.install(e[0], e[1]);
});
```

package.json
```diff
@@ -52,6 +52,7 @@
     "promise": "6.x.x",
     "request": "2.x.x",
     "specberus": "1.x.x",
+    "mkdirp": "0.x.x",
```
astorije: To match the other versions, this should be `0.5.x` instead.
Would that be a problem for you?

deniak: Done in a0d9f58beeecf4063ff14be07ef8dd1855a3197b

astorije:
*A posteriori* :+1: :-)

PR #131, opened 2015-03-17
Make it clear that Eds. don't need to install this (closes #130)

marcoscaceres:

PR #133, opened 2015-03-18
Make the README clearer regarding what should editors and developers read to use the tool

astorije:
Extending #131 hoping to be clearer and more explicit.


tripu:
@astorije, I took the liberty of committing tiny changes of style before merging&hellip;

astorije:
@tripu, instead of adding commits like this and merge right away, please ask me to do the changes, or wait for my :+1: to merge if you really want to commit on top of my branch (first solution preferred in most FOSS projects). As you can imagine, there is probably a good reason why I do a change if it's not a typo. Thanks!

PR #136, opened 2015-03-20
Get rid of the code-surrounded links and fix a typo

astorije:
See #133


tripu:
My edition in #133 wasn't accidental: URLs were displayed using a monospace font before; I think technical documents look better and are more usable that way. But no big deal :¬)

PR #138, opened 2015-03-26
Fix #137: Fetch files as Buffer by default instead of utf8 strings

### LINKED ISSUE #137: Echidna appears to corrupt images (probably all binary files)
If you compare images as available in the original source (e.g. http://www.w3.org/html/wg/drafts/html/wd/img/sample-details-1.png) with those available in the installed version (the same would be http://www.w3.org/TR/html51/img/sample-details-1.png) you will note some differences.

The most glaring one is that the latter does not work. Running the new file through `file`, it no longer is recognised as PNG. The size has been hugely inflated: 337K instead of 181K.

I haven't looked at the code but the most likely source for this problem is that at some point binary files get treated as text. All the files — including textual ones — should be treated as opaque binaries by this tool chain

astorije:


astorije:
@tripu, I assigned @darobin here because we discussed together about this bug and I could use his advice if I got it wrong :-)

It should deserve its own patch version, I think.

lib/document-downloader.js
```diff
@@ -62,8 +65,10 @@ DocumentDownloader.fetchAndInstall = function (url, dest) {
     if (!pathExists) return mkdir(dest);
   }).then(function () {
     return DocumentDownloader.fetch(url).then(function (content) {
-      if (content.trim().charAt(0) !== '<') {
-        var filenames = DocumentDownloader.getFilenames(content);
+      var contentUtf8 = content.toString('utf8');
+
+      if (contentUtf8.toString('utf8').trim().charAt(0) !== '<') {
+        var filenames = DocumentDownloader.getFilenames(contentUtf8);
```
darobin: You're converting to UTF-8 twice, probably harmless but not needed.

darobin:
Apart from the double `toString()` this LGTM. Please merge once that bit is fixed.

Should we have a "hotfix" label to mark things that need to be deployed quickly?

astorije:
Ah ah right, misused copy & paste!

Fixed in 5a635e6, I will merge as soon as Travis CI completes.

@tripu, okay to tag this as a patch version and release it as soon as you see this message?
Also, what about @darobin's suggestion regarding a **hotfix** label? My opinion is actually to have a **bugfix** label instead, and to deploy whenever a PR with this label is merged, because I think all bugfixes will be urgent for someone. Does that make sense?

tripu:
We already have a *bug* label (just added to this PR). I'm not sure we need more granularity to rank bugs and their fixes. I'd say most bugfixes are *not* urgent, really.

But this will be included in the next version deployed, of course.

darobin:
Well, ideally the code would be deployed whenever something is merged to master :)

But if we can't have that, then I would suggest that the ability to flag some issues as requiring quicker attention would be helpful. This bug for instance has been causing the HTML specification (and presumably any other spec that uses images) to be broken for over two weeks. We've already had people report it several times, it's getting to be a bit embarrassing really :-/

astorije:
> Well, ideally the code would be deployed whenever something is merged to master :)

Don't open that box, @darobin! :-)

@tripu, I agree with @darobin regarding this issue: this one was really, really urgent and would have been deployed super quickly. That said, I haven't deployed Echidna yet so I don't know what are the steps to do so, but I am generally in favor of [one-step builds](http://www.joelonsoftware.com/articles/fog0000000043.html) when automatic deployment is not possible or not desired. If it's that easy, then developers should be able to deploy patches like this one.

PR #143, opened 2015-04-10
source checker: authorized whether we use http or https

deniak:


astorije:
Shouldn't we favor HTTPS instead? Like authorize HTTPS escalations but not HTTP... *descalation*? :-)

It just seems odd to me that we blindly switch among them when it's the role of the web server on the other side. Remember that a same URL can be set to serve something really different for these 2 schemes.

tripu:
I guess you're right, @astorije. I too thought about being more conservative, allowing changes only in one way instead (what you say about &ldquo;escalation&rdquo;).

I merged as is for the sake of simplicity and to avoid delaying it.

(I don't think this is going to cause problems in the real world. We are not comparing URLs verbatim, just URL &ldquo;origins&rdquo;. Your warning about content potentially being different over different protocols applies also to variations over the same base URL, which we actually allow on purpose.)

PR #144, opened 2015-04-10
Make domains match github.io instead of github.com

astorije:


config.js.example
```diff
@@ -22,5 +22,5 @@ global.ALLOWED_CLIENTS = [
   // W3C:
   /https?:\/\/(www\.)?w3c?\.org/i,
   // GitHub:
-  /https?:\/\/github\.com/i
+  /https?:\/\/.*\.github\.io/i
```
tripu: This would let `http://malicious.com/github.io/boom` pass.
I suggest `/https?:\/\/[^.]+\.github\.io/i`

tripu: (Come to think of it, and for the same reason, all these regex's should match the beginning of line too, eg `/^http…`. But, no! because in [`app.js#L120`](https://github.com/w3c/echidna/blob/master/app.js#L120) these are used to test *the origin header only*, and that shouldn't contain directories in it&hellip; So I think we're safe with this, @astorije. I'll merge tomorrow if you don't say otherwise.)

astorije: I must admit I haven't digged into the CORS part of Echidna's code so far. Giving it a look now, I'm  confused (and tired :-) ) if we are safe or not. I'd be more comfortable if someone else than me can give a review before merging (plus there would be a conflict of interest :p), but decision to merge comes down to you so it's your call!

I guess a better solution would be to write some tests around the CORS snippet of code, if you are feeling bored... :p

tripu:
This is good enough.

darobin:
Note: I don't know if this is intended or not, but absolutely anyone can procure a `github.io` address. At that point, I wonder if it really makes sense to have CORS at all?

astorije:
Huh @darobin?

darobin:
CORS prevents arbitrary web sites from accessing the service in the client. `github.io` allows the creation of pretty arbitrary sites in the client. Allowing `github.io` means allowing from pretty much everyone.

tripu:
Right, @darobin. What about enabling specific repos (subdomains) instead, making the CORS filtering more granular?

So, the list of allowed regex'es wouldn't read

```javascript
global.ALLOWED_CLIENTS = [
  /https?:\/\/localhost(:\d{1,4})?/i,
  /https?:\/\/(www\.)?w3c?\.org/i,
  /https?:\/\/.*\.github\.io/i
];
```

but more like

```javascript
global.ALLOWED_CLIENTS = [
  /https?:\/\/localhost(:\d{1,4})?/i,
  /https?:\/\/(www\.)?w3c?\.org/i,
  /https?:\/\/w3c\.github\.io/i,
  /https?:\/\/whatwg\.github\.io/i,
  /https?:\/\/webplatform\.github\.io/i,
  /https?:\/\/awesome-collaborator\.name/i,
  /https?:\/\/respected\.contributor\.com/i
];
```

I know, I know&nbsp;&mdash;&nbsp;this array is going to be a pain to manage if each one of us start publishing from our own home-brewed client, deployed in a different domain. But if that happens, it'll be a sign that Echidna is a huge success, and then I'll be happy to refactor this simple CORS mechanism :¬)

darobin:
I would suggest:

* If we believe that there is a threat model here, then we should enable this only from `w3c.github.io` (for now) and add more people later as you suggest. (Maybe add @marcoscaceres right now so he stops complaining ;)
* If we don't see a threat, then just have CORS set to `*`.

So... what's the threat? :)

Issue #148, opened 2015-04-17
Consolidate request status strings


astorije:
Following the discussion on [a comment on #84](https://github.com/w3c/echidna/pull/84/files#r28552721):

```js
var STATUS_STARTED = 'started';
```

should be consolidated in the `RequestState` class.

PR #154, opened 2015-04-17
Get rid of request status strings

astorije:


app.js
```diff
@@ -101,7 +98,7 @@ app.post('/api/request', function (req, res) {
       'version-specberus': SpecberusWrapper.version,
       decision: decision,
       state: new RequestState(
-        STATUS_STARTED,
+        '',
```
astorije: I know, this is not pretty, but it will evolve: things will move from `app.js` to `lib` components...

tripu:
I'm not sure I understand what you want to achieve here, @astorije. One thought:

You removed pseudo-constants `STATUS_ERROR` and `STATUS_SUCCESS` from the orchestrator and replaced them with string literals. Isn't that the opposite of what we want (#148)? Shouldn't all those *status* values be defined only in `request-state.js`?

Apart from that doubt, another idea:

The constructor of `RequestState` is used only once outside the class itself: in `app.js`, always with an empty string as the first argument. I agree with you that it doesn't look nice. Maybe we can remove that first parameter altogether, and set that initial status (which is always the same) within the constructor itself? (It is `''` now; I suggest an explicit string, like `'STATUS_CREATED'`, `'STATUS_INIT'` or something like that.) Then, the method `RequestState.set()` can set the `status` of the newly created object accessing that property directly, after the constructor.

astorije:
> I'm not sure I understand what you want to achieve here, @astorije. One thought:

In a nutshell, simplify things.

> You removed pseudo-constants `STATUS_ERROR` and `STATUS_SUCCESS` from the orchestrator and replaced them with string literals. Isn't that the opposite of what we want (#148)? Shouldn't all those *status* values be defined only in `request-state.js`?

Indeed, because there is no point, at the moment, in having `RequestState.something = 'something';` with `Request.something` used exactly once after that (again, at the moment). That's one extra level of indirection that is just making noise. The idea of #148 is that `STATUS_*` were defined in two places, now it's not anymore :-)
Again, it will evolve, but right now I go for simpler.

> The constructor of `RequestState` is used only once outside the class itself: in `app.js`, always with an empty string as the first argument. I agree with you that it doesn't look nice. Maybe we can remove that first parameter altogether, and set that initial status (which is always the same) within the constructor itself? (It is `''` now; I suggest an explicit string, like `'STATUS_CREATED'`, `'STATUS_INIT'` or something like that.) Then, the method `RequestState.set()` can set the `status` of the newly created object accessing that property directly, after it creates it.

Twice: in `app.js` and in tests :-)
Regarding the constructor, I'm not too fond of the idea. As you notice, `Job` and `RequestState` classes are very similar (others will too) and will be abstracted when I find the nicest way to do it, most likely with input from @tabatkins since we discussed about it on IRC.
I suggest leaving things simple right now and bet on a nice abstraction sooner or later.

astorije:
Just rebased to make the merging possible.

tripu:
> Indeed, because there is no point, at the moment, in having `RequestState.something = 'something';` with `Request.something` used exactly once after that (again, at the moment). That's one extra level of indirection that is just making noise. The idea of #148 is that `STATUS_*` were defined in two places, now it's not anymore :-) Again, it will evolve, but right now I go for simpler.

I personally like to avoid all literals in the middle of the code (string literals more than anything else) and hoist them as pseudo-constants. Even if they are used just once.

>> The constructor of `RequestState` is used only once outside the class itself.

> Twice: in `app.js` and in tests :-)

Where in the tests?

astorije:
> I personally like to avoid all literals in the middle of the code (string literals more than anything else) and hoist them as pseudo-constants. Even if they are used just once.

Because... ? :-) I'm open to arguments.
Mines are simplicity, no over-engineering, open to future abstraction.

> Where in the tests?

[Here](https://github.com/w3c/echidna/blob/57f29333784e0c0dddb83db5a52fd0f51f54b445/test/test-orchestrator.js#L17) and [there](https://github.com/w3c/echidna/blob/57f29333784e0c0dddb83db5a52fd0f51f54b445/test/test-request-state.js#L16) for example (okay, the second one is arguable). The key point here is that it makes more sense to keep things consistent for a (near) future abstraction.

tripu:
> Because... ? :-) I'm open to arguments.
> Mines are simplicity, no over-engineering, open to future abstraction.

I think most literals and magic numbers scattered through the code are error-prone, so I try to avoid them as much as I can, or at least declare them only once and keep those declarations confined to specific places. (That with a few exceptions, like zeroes, empty strings etc, which are so handy, and less likely to be got wrong by mistake, anyway.) Among those, string literals are the worse, IMHO.

Besides, almost every time you need a string literal you will need it more than once (otherwise, why to use precisely that literal, and not something else?). So having those declared as [pseudo-]constants in the first place proves useful in the long run, I think&hellip;

Long rant, no big deal. Happy to leave it as it is :¬)

> [Here](https://github.com/w3c/echidna/blob/57f29333784e0c0dddb83db5a52fd0f51f54b445/test/test-orchestrator.js#L17) and [there](https://github.com/w3c/echidna/blob/57f29333784e0c0dddb83db5a52fd0f51f54b445/test/test-request-state.js#L16) for example (okay, the second one is arguable).

(Ah, right. Sorry, I didn't find those in `master` since that one's not merged yet.)

Issue #147, opened 2015-04-14
Break down tests


astorije:
That's a big test file that we start to have here!

It would be nice to break it down into different files so that for every `lib/[name].js` file, there is a corresponding `test/[name]-suite.js` test suite.

Mocha lets us have multiple `*.js` files (but we have to augment the `.jshintignore` file until flagging these files as test suites) and will run against each of them in alphabetical order, but we might want to export the common pieces (`chai` loading, ...) to a shared file, [maybe with `require`](http://stackoverflow.com/a/20723913/1935861). Or not, I don't know what is best here yet, like, will doing so prevent us to run tests in a standalone way, ...?

astorije:
@plehegar, I just tried to extract tests related to the Document Downloader into its very own file `test-document-downloader.js`. It is not possible as is, because the line:

```js
var server = require('./lib/testserver');
```

would exist in multiple files, each starting its own test server, which collides with other things (and even if it didn't, that's not cool).

Could you look into this to start the test server only once and get a hold onto it after that?

PR #156, opened 2015-04-21
Add tests to PR #84

astorije:
64 tests before, 112 tests after (**48 new tests**). Most of them are dead simple but that will help detect issues when refactoring small or big portions.

The scary part is the code coverage: it is marked as **98%** at the moment, while it will go down to **62%** when this PR is merged. However, do not get fooled, the overall coverage *does* increases with this PR.
This comes from the fact that this PR asks [Istanbul](https://github.com/gotwarlost/istanbul) to cover all source files where it used to take into account only files that were traversed by the tests. It means that source files that were absolutely not tested... were not counted! Not very accurate.

A question for @tripu: should we add `app.js` to the files ignored by the coverage? On one hand it seems like cheating, on the other hand we will *never* be able to test it using [Mocha](http://mochajs.org/) and our test suite...

### DIFF
--- a/.jshintignore
+++ b/.jshintignore
@@ -6,3 +6,4 @@ Gruntfile.js
 config.js
 test/views
 test/test.js
+test/test-*.js
--- a/.istanbul.yml
+++ b/.istanbul.yml
@@ -1,0 +1,5 @@
+instrumentation:
+  include-all-sources: true
+  excludes:
+    - "**/bower_components/**"
+    - "Gruntfile.js"
--- a/functions.js
+++ b/functions.js
@@ -7,7 +7,7 @@ var Map = require('immutable').Map;
 var Specberus = require('specberus/lib/validator').Specberus;
 require('./config.js');
 
-var SpecberusWrapper = function () {};
+var SpecberusWrapper = {};
 
 SpecberusWrapper.validate = function (url) {
   return new Promise(function (resolve, reject) {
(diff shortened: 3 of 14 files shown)


.jshintignore
```diff
@@ -6,3 +6,4 @@ Gruntfile.js
 config.js
 test/views
 test/test.js
```
astorije: This line will be removed when tests are properly split as per #147.

astorije:
@tripu, @darobin, this PR doesn't affect code, just tests, and the ones who failed in the last Travis CI test are untouched... Do you know if something changed on the Specberus side... or maybe the CSS validator side?

tripu:
@astorije, see #159. It's a PR over this other branch of yours (sorry, I just realised it probably should be a normal PR into `master`&hellip;).

If you merge that one, and that fixes the build over here (as I expect), I'll merge this PR in turn :¬)

astorije:
@tripu, I'd say this PR can now be merged. I updated the original description to give you some food for thoughts.

astorije:
@tripu? Pleaaaase :-)
`master` will be broken until merging :-(

astorije:
\o/ Thanks tripu!

tripu:
About code coverage: @astorije, I'd say we include as much as we can&nbsp;&mdash;&nbsp;even code that is impractical or impossible to test. That's the only way for the metric to be really meaningful, isn't it? Better not to fool ourselves with distorted figures. Just an idea :¬)

PR #159, opened 2015-04-23
Update expected errors in tests, post w3c/specberus#173

tripu:
w3c/specberus#173 added a few new possible errors; among them, [`latest-is-not-previous`](https://github.com/w3c/specberus/pull/173/files#diff-09f641a52baf89d6d1785acc66dafe97R172).

These three drafts used by the test suite are affected:

* `navigation-timing-2`
* `nav-csserror`
* `nav-csswarning`

This PR updates (adds one to) the total count of errors expected by mocha for these three test specs, thus fixing the result of the test suite.

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -371,7 +371,7 @@ describe('SpecberusWrapper', function () {
 
     it('should return an error property that is an empty list', function () {
       return expect(content).that.eventually.has.property('errors')
-        .that.is.empty;
+        .that.has.size(1);
     });
 
     it('should promise an object with a metadata property', function () {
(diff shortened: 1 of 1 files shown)


astorije:
@tripu, I merged this one and was willing to update the labels of the tests ("has no errors" [...] `has.size(1)`), but then realized that this might "break" the tests as @plehegar intended them when he wrote them at first.

@plehegar, would you mind taking a look at these tests and see if they need to be changed furthermore, like changing the drafts instead of the tests, adding n tests, ... ? Also, right now this is not merged into `master` but will be soon.

PR #161, opened 2015-04-24
Fix #160: Make sure filenames and destination files lists match after sanitizing

### LINKED ISSUE #160: Manifests do not get sanitized properly
At the moment, if a manifest contains lines to be filtered out at sanitization time, the downloading will be completely mangled: files will have wrong filenames and it will eventually fail.

The reason is because only one of the 2 lists built from the manifest (the list of URLs to download, and the list of files to install on the disk) gets sanitized, which is a bug.

astorije:
The bug fix is essentially at [line 78](https://github.com/w3c/echidna/pull/161/files#diff-bac32d994d58faa28bb71801bbf1de35R78), but I cleaned up some other things around and added a docstring (yes, just for one function of the file :p).

### DIFF
--- a/lib/document-downloader.js
+++ b/lib/document-downloader.js
@@ -48,15 +48,20 @@ DocumentDownloader.installAll = function (destsContents) {
   }));
 };
 
-DocumentDownloader.sanitize = function (list) {
-  return list.filter(function (filename) {
-    if (filename.toLowerCase().indexOf('.htaccess') !== -1) return false;
-    else if (filename.toLowerCase().indexOf('.php') !== -1) return false;
-    else return true;
-  });
+/**
+ * Checks if a given filename is allowed to be published by the system.
+ *
+ * @param {string} filename - The filename to check against
+ * @returns {boolean} `true` if the filename is allowed, `false` otherwise
+ */
+DocumentDownloader.isAllowed = function isAllowed(filename) {
+  if (filename.toLowerCase().indexOf('.htaccess') !== -1) return false;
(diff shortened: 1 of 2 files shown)


astorije:
Oh by the way, the build will fail because of #159 not being merged to master but to a non-merged branch yet (see #156)...

tripu:
@astorije, is this `_` alias for the class itself some common practice? I don't remember seeing it, and I'm not sure I like it.

Merging nonetheless.

astorije:
@tripu, I understand your concern, but think of it that way: `fetch()`, `fetchAll()`, `install()`, `installAll()`, `getFilenames()`, `isAllowed()` are all static methods that must be prefixed with `DocumentDownloader` as, in JS, there is no `self::` meta-variable and it's not possible to call them without the class name.
So I am using `_` as an anonymous variable (which is actually a *fairly* common practice in programming languages, to use `_` as an anonymous or mute variable).
Additionally, it's a local variable so nothing that will leak to the API (as in code API, not JSON API) :-)

tripu:
Understood, @astorije.

I guess in a previous life I got used to seeing `LongClassName.staticMethod()` all the time, eg in Java, so it no longer hurts my eyes :¬) Never liked the look of all those `_` and `__` that Python is so fond of, for example&hellip;

I understand that when it's a common practice, the meaning is clear for everyone, and so there is no problem. Still, it's funny how all that runs counter to the general principle &ldquo;be verbose and give your variables meaningful names&rdquo;.

PR #165, opened 2015-04-28
Mail notifications: fix one bug, and improve subject lines and messages

tripu:
(I know&hellip; these are different issues, and should be different PRs. I use `tripu/miscellanea` for bunches of trivial fixes that usually can be safely processed together. It was only later that I saw that was not the case this time. If anything here is controversial, I'm happy to split into different PRs, of course :¬)

* Fix [bug](https://github.com/w3c/echidna/commit/36c3fc75f2173a8d2b7fb8052ed31bec8ef63369#diff-0364f57fbff2fabbe941ed20c328ef1aR147): missing space between command-line arguments sent to the *sendmail* script.
* Two improvements to notifications sent to the mailing list:
  * Subject line more compact, easier to parse quickly when browsing mail archives, or when only the beginnings of subject lines are visible, eg on a narrow mail client:
  ```
[W3C Publication] Success: <URL>
[W3C Publication] Failed: <URL>
[W3C Publication] Success: <URL>
[W3C Publication] Success: <URL>
[W3C Publication] Failed: <URL>
[W3C Publication] Failed: <URL>
[W3C Publication] Success: <URL>
[W3C Publication] Success: <URL>
[W3C Publication] Success: <URL>
[W3C Publication] Failed: <URL>
```
  * For successful publications, include versions of Echidna and Specberus too. This means that the 4th argument of the script is mandatory now, too.
* Added notes for contributors about which versions of Node.js and nps we use now.

### DIFF
--- a/app.js
+++ b/app.js
@@ -142,7 +142,8 @@ app.post('/api/request', function (req, res) {
         cmd += ' \'' + JSON.stringify(state, null, 2) + '\'';
       }
       else {
-        cmd += ' \'Echidna ' + meta.version + '; Specberus ' + SpecberusWrapper.version + '\'';
+        cmd += ' \'Echidna ' + meta.version +
+          '; Specberus ' + SpecberusWrapper.version + '\'';
       }
 
       console.log('[' + state.get('status').toUpperCase() + '] ' + url);
--- a/app.js
+++ b/app.js
@@ -141,6 +141,9 @@ app.post('/api/request', function (req, res) {
       if (state.get('status') === 'error') {
         cmd += ' \'' + JSON.stringify(state, null, 2) + '\'';
       }
+      else {
+        cmd += ' \'Echidna ' + meta.version + '; Specberus ' + SpecberusWrapper.version + '\'';
+      }
 
       console.log('[' + state.get('status').toUpperCase() + '] ' + url);
       exec(cmd, function (err, _, stderr) { if (err) console.error(stderr); });
--- a/app.js
+++ b/app.js
@@ -1,4 +1,3 @@
-
 'use strict';
 
 console.log('Launching…');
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by tripu** [comment]
I don't understand why this build fails. I restarted it a couple of times, with the same result&hellip;

:¬?

File: app.js [hunk 1]
```diff
@@ -1,3 +1,4 @@
+
```
astorije: No you didn't! :-)

app.js
```diff
@@ -135,11 +136,15 @@ app.post('/api/request', function (req, res) {
       requests[id].state
     ).then(function (state) {
       var cmd = global.SENDMAIL + ' ' + state.get('status').toUpperCase() +
-        ' ' + global.MAILING_LIST + url;
+        ' ' + global.MAILING_LIST + ' ' + url;
```
astorije: Sorry again about that :-) Thanks for spotting it!

astorije:
@tripu, I don't know if you have resolved your failing build or not. The first failure is a JSCS issue (that you fixed) and the second one is because coverage decreased a little bit, nothing to worry about I think.

astorije:
So, for this PR, or PR-all-in-one :-) : 

- :-1: for blank lines at top and bottom of files (unless it's a line feed at bottom, which are not the case here). Please remove.
- Undecided regarding the paragraph on `node` and `npm` versions as we discussed on another thread, but :-1: to duplicate the statements. It should be on the README file **or** the CONTRIBUTING file (**or** the wiki), but not both of them. Please remove one of them. As for the text itself, we need resolution on the other thread.
- :+1: on fixing the command line call
- :+1: on improving the email sent to the mailing-list (the syntax could be better but it's good as is at the moment)

tripu:
@astorije, I edited a few things according to your comments. If you like it now, please merge.

I left that note only in `CONTRIBUTING`. We can continue discussing what versions to use in productions, and they may change in the near future. In the meantime, the note is accurate, so I suggest let's leave it as it is.

astorije:
Thanks, looks good to me! :+1:

PR #170, opened 2015-05-05
Fix #162: Remodel dependency versions and avoid insecure express package (and also, fix wrong version of JSCS...)

### LINKED ISSUE #162: Solve the "dependencies: insecure" badge and re-think dependencies strategy
In #50, I suggested to use format `N.x.x` with `N` fixed to use any version as long as the backward-compatibility was ensured.
However, I was expecting `N.x.x` to mean "the latest version of the `N`-th major version" but it actually means "Any version that has `N` as the major version". In practice, it means that a system having a package in version `N.0.0` globally installed will not have this package updated (or re-installed locally) to the latest version when running `npm install`, even if we have been using features introduced in, say, version `N.5.0`. That will make Echidna fail until the package is removed from the system or globally updated.

This issue also explains why [David is reporting an *insecure* dependency](https://david-dm.org/w3c/echidna) as our definition in `package.json` says that Echidna will work with `express` version `4` to `4.5`, which are deemed insecure.

To solve this, we can change our dependency version strategy to `~`, `^` or `>` (see [`npm help package.json`, section `dependencies`](https://docs.npmjs.com/files/package.json#dependencies) and [`semver(7)`](http://manpages.ubuntu.com/manpages/precise/man1/semver.1.html)). I haven't looked into pros and cons for each descriptor. Any preference?

astorije:


astorije:
Thanks @tripu! It does seem like a bigger trouble when updating the dependencies, but actually it gives something to compare with when checking for incompatibilities or new features and that will be a priceless time saver!

Example: when adding JSCS rules recently (#167), I had to check when JSCS was introduced in our commit history and compare the date of the commit with the date of the according release of JSCS... urgh! Easier now. I hope this makes sense.

Issue #97, opened 2015-02-23
Make Echidna run locally without failing


astorije:
Although Echidna is supposed to be run locally, there are parts of the system that currently fail if you do not have a full W3C setup. This issue is meant to keep track of what needs to be done and what it already done to fix that.

Here are a list of the steps with their status regarding ability to run locally:
- [x] **Document Downloader**: does not interact with W3C systems
- [ ] **Specberus**: works as a standalone project and as a Node module, but still markup/css validator issue
- [x] **Token Checker**: a mock API is provided
- [x] **Third Party Checker**: works as a standalone project and as a Node module
- [x] **W3C DB Publisher**: a mock API is provided
- [x] **TR Installer**: shares interface with `cp -r` so it can be mocked locally
- [x] **Shortlink Updater**: shamefully use `#` as a local command... we can probably do better, but at least it works...
- [ ] Serve the staging documents through a webserver started with Node.js rather than expect the configuration to point to a server set up somewhere else

astorije:
I should actually use the [issue template](https://github.com/w3c/echidna/blob/master/CONTRIBUTING.md#report-a-bug-or-suggest-a-feature-idea) for this.
@anssiko or someone else, care to run it locally to share the current output? Thanks!

astorije:
@anssiko, I am currently working on this one, it should be ready soon.

anssiko:
@astorije awesome, let me know when I can give it a spin.

astorije:
@anssiko actually, you already can, but there are still things that need to be fixed. Here is what you have to do:

- Somewhere on your machine, `git clone https://github.com/w3c/echidna.git && cd echidna && git checkout astorije/make-it-standalone`
- `npm install` to install the dependencies
- `cp config.js.example config.js` to create the configuration file and then edit `global.DEFAULT_TEMP_LOCATION` and `global.DEFAULT_RESULT_LOCATION` to directories the webserver can write (I use `/home/jeremie/public_html/tr`), and `global.DEFAULT_HTTP_LOCATION` to a web-served location (I use `http://localhost/~jeremie/tr`)
- I actually broke Echidna in a previous commit, sorry, so until #169 is merged, you need to manually apply the fix with `git merge astorije/fix-orchestrator`
[...]

PR #171, opened 2015-05-06
Make Echidna run standalone

astorije:
See #97.

WIP, do not merge.


astorije:
@tripu, I just rebased to make this branch mergeable.
I know #97 is not completely closed with this, but this PR already adds a lot of things and that would be nice to merge it. The last 2 bullet points that need solving are in the end independent of this PR.

Anyway, you can start the review now, give your feedback, and I'll make the changes you'll request so that we can merge it. I'll deal with the other points later as there are more urgent things unfortunately. Plus, this PR still makes Echidna a lot more standalone than it used to be, so it's always a plus.

Let me know if you have any question.

config.js.example
```diff
@@ -6,14 +6,14 @@ global.DEFAULT_LOG_LOCATION = '.';
...
-global.W3C_PUBSYSTEM_URL = '';
-global.TR_INSTALL_CMD = '/bin/cp -R';
-global.UPDATE_TR_SHORTLINK_CMD = '';
-global.SENDMAIL = 'sendmail.sh';
+global.USERNAME = '';
+global.PASSWORD = '';
+global.W3C_PUBSYSTEM_URL = 'http://localhost:3001/publish';
+global.TR_INSTALL_CMD = 'mkdir -p $dest && cp -R $source $dest';
+global.UPDATE_TR_SHORTLINK_CMD = '#';
+global.SENDMAIL = './sendmail.sh';
```
tripu: Ah. Good catch. I just checked in production to see how this could possibly work&nbsp;&mdash;&nbsp;when we deploy there, we edit it to have the absolute path to the script.

tripu:
Good job, @astorije!

PR #176, opened 2015-05-15
Fix express directory for static assets

tripu:
Because `NODE_ENV` is set in the code only [in `test/test.js`](https://github.com/w3c/echidna/blob/master/test/test.js#L4) (to `'dev'`), the expression `process.env.NODE_ENV === 'production'` was always `false`. Thus the first clause of the condition never executed¹.

Apart from that, [the default value of the app setting `view` is `process.cwd() + '/views'`](http://expressjs.com/4x/api.html#app.set), so there is no need to set it manually. Only the directory for static content `assets/` needs to be set, as described [here](http://expressjs.com/starter/static-files.html). No need to use `dist/`directories either as we're not using Grunt tasks for configuration management.

-- 
¹ `NODE_ENV` is still set to `production` manually in the command line when deploying, as described in the [internal docs](https://www.w3.org/Systems/Publishing/new-workflow#start).

### DIFF
--- a/app.js
+++ b/app.js
@@ -30,17 +30,7 @@ var argResultLocation = process.argv[5] || global.DEFAULT_RESULT_LOCATION;
 app.use(compression());
 app.use(bodyParser.urlencoded({ extended: false }));
 app.use(corsHandler);
-app.use(express.static('views'));
 app.use(express.static('assets'));
-
-if (process.env.NODE_ENV === 'production') {
-  app.set('views', __dirname + '/dist/views');
-  app.use(express.static(__dirname + '/dist/assets'));
-}
-else {
-  app.set('views', __dirname + '/views');
-  app.use(express.static(__dirname + '/assets'));
-}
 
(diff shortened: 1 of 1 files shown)


deniak:
@tripu LGTM.
However, can't we remove the [`grunt` dependencies](https://github.com/w3c/echidna/blob/master/package.json#L63) as well?

tripu:
I think so, @deniak. [I restored those dependencies at some point](https://github.com/w3c/echidna/pull/71), but now I think we should prune, [like @astorije did before](https://github.com/w3c/echidna/commit/28aa84c140121c3eb037eb2f6edc42dd1cbf9651).

Either we invest the time to justify the use of Grunt, set it up properly and document, or we get rid of it for simplicity in the meantime. I prefer option no. 2 right now.

@deniak, maybe merge this one? I'll submit a different PR regarding Grunt, and assign to @astorije to get his comments.

PR #177, opened 2015-05-17
Remove unused configuration lines

astorije:
`global.PHANTOM` was most likely made useless [when `phantomjs` was added as a dependency of the third-party-resources-checker](https://github.com/w3c/third-party-resources-checker/commit/59d0f26aadcbbafb07e6820e45d086ceef60a8a8#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R12) a while ago.

Regarding `global.RESOURCES_WHITELIST`, I'm afraid I can't even find in the history when it was last used. Either not using it is a mistake or is because we don't need to whitelist anything [more than the default whitelist](https://github.com/w3c/third-party-resources-checker/blob/2431a578ca8a5e3cbeed9cc59494c874e3bca5cd/lib/detect-phantom.js#L19). Either way it needs flagging.


deniak:
> `global.PHANTOM` was most likely made useless [when `phantomjs` was added as a dependency of the third-party-resources-checker](https://github.com/w3c/third-party-resources-checker/commit/59d0f26aadcbbafb07e6820e45d086ceef60a8a8#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R12) a while ago.

+ 1

> Regarding `global.RESOURCES_WHITELIST`, I'm afraid I can't even find in the history when it was last used. Either not using it is a mistake or is because we don't need to whitelist anything [more than the default whitelist](https://github.com/w3c/third-party-resources-checker/blob/2431a578ca8a5e3cbeed9cc59494c874e3bca5cd/lib/detect-phantom.js#L19). Either way it needs flagging.

Regarding the whitelist, it's needed but currently not used. The hardcoded whitelist only allows resources coming from `www.w3.org`. In the future, we might want to open it to other subdomains or allows external scripts.

astorije:
> Regarding the whitelist, it's needed but currently not used. The hardcoded whitelist only allows resources coming from `www.w3.org`. In the future, we might want to open it to other subdomains or allows external scripts.

That's what I understood, that the default whitelist was enough for our needs.
However, you say it's needed, how? If it's not used, I'm not sure it's needed.
Anyway, if it's only for future purposes, then we can safely remove it until such time comes.

deniak:
> However, you say it's needed, how? If it's not used, I'm not sure it's needed.
> Anyway, if it's only for future purposes, then we can safely remove it until such time comes.

If someone wants a specific resource to be whitelisted, it's more convenient to simply add it to the list (currently empty) than creating a new MR and wait for it to be deployed. Also, I think that whitelist file should be under github.
Since it's already here and we'll need it, I'd prefer to keep it.

astorije:
> If someone wants a specific resource to be whitelisted, it's more convenient to simply add it to the list (currently empty) than creating a new MR and wait for it to be deployed.

Well yeah, except that this whitelist is not used anywhere... So I'll re-add the line in the config file and call it [here](https://github.com/w3c/echidna/blob/01a01f479505a96dfcc8ec0ae7dd2e45653392e7/lib/orchestrator.js#L216).
By the way, if the whitelist is under GitHub, then we will have to go through PR process and deployment, or use a temporary hacked whitelist in production, which is never recommended...

astorije:
@deniak I added what we discussed about. The line in `orchestrator.js` is a bit clumsy but that's the easiest at that point (especially considering this function(s) is not tested anyway...).

astorije:
Hum, Travis didn't trigger, odd...

PR #178, opened 2015-05-18
Remove all traces of Bower and Grunt

tripu:
&hellip;until we decide to start using either of them properly.

--
See https://github.com/w3c/echidna/pull/176#issuecomment-102940191


astorije:
:+1: Yes! This makes me happy on a Monday morning \o/

PR #179, opened 2015-05-18
Avoid CDN dependencies and upgrade jQuery

tripu:
* jQuery upgraded to latest stable release.
* JSHint exclusions more fine-grained to avoid accidental ignores; its lines sorted alphabetically.
* Web dependencies load now from the app web space itself (ie, not from a CDN); in both the main app and the testserver UI.


astorije:
Goodbye CDN \o/

:+1:

PR #181, opened 2015-05-19
HTML markup cleanup

astorije:
- Removed the trailing slashes
- Removed the unnecessary `type` attribute for `style`, `link` and `script` (see [this](http://stackoverflow.com/a/6239035/1935861))
- Removed unnecessary `shortcut` value for `rel` attribute of the favicon

### DIFF
--- a/test/views/index.html
+++ b/test/views/index.html
@@ -1,32 +1,28 @@
 <!DOCTYPE html>
-
 <html lang="en-GB">
-
     <head>
-        <meta charset="utf-8" />
-        <meta name="application-name" content="Echidna testbed" />
-        <meta name="description" content="W3C's new publication workflow testbed page" />
-        <meta name="keywords" content="w3c,publication,workflow,echidna,test" />
-        <meta name="author" content="W3C" />
+        <meta charset="utf-8">
+        <meta name="application-name" content="Echidna testbed">
+        <meta name="description" content="W3C's new publication workflow testbed page">
+        <meta name="keywords" content="w3c,publication,workflow,echidna,test">
+        <meta name="author" content="W3C">
         <title>Echidna testbed</title>
-        <link rel="stylesheet" type="text/css" href="style.css" />
+        <link rel="stylesheet" href="style.css">
     </head>
-
     <body>
-
         <h1>Echidna testbed</h1>
 
         <h2>Settings</h2>
 
         <form class="left">
-            <label>Host<br /><input id="host" type="text" value="http://localhost" autofocus /></label>
-            <label>Port<br /><input id="port" type="number" value="3000" min="0" max="9999" step="1" /></label>
-            <label>Endpoint<br /><input id="suffix" type="text" value="/api" /></label>
+            <label>Host<br><input id="host" type="text" value="http://localhost" autofocus></label>
+            <label>Port<br><input id="port" type="number" value="3000" min="0" max="9999" step="1"></label>
+            <label>Endpoint<br><input id="suffix" type="text" value="/api"></label>
         </form>
 
         <form class="right hidden">
-            <label>Refresh rate (<em>s</em>)<br /><input id="rate" type="number" min="0.1" max="60" step="0.5" value="1.0" /></label>
-            <label>Auto-scroll<br /><input id="scroll" type="checkbox" checked /></label>
+            <label>Refresh rate (<em>s</em>)<br><input id="rate" type="number" min="0.1" max="60" step="0.5" value="1.0"></label>
+            <label>Auto-scroll<br><input id="scroll" type="checkbox" checked></label>
         </form>
 
         <h2>Specs</h2>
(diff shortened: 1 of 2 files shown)


tripu:
What is the rationale for removing all those blank lines?

* They visually separate major sections, eg *head* and *body*.
* Blank lines improve readability too around groups of related elements, eg to visually *isolate* a bunch of *script* elements from what's before and after them.
* At the end of files, they leave some margin in text editors, and clearly mark the EOF in the command line when `cat`ting etc.
* If I had to logically split an HTML5 document in two parts, for me it would be: the *doctype*, and everything else (the document itself). If there's one place where some air, or padding, seems appropriate semantically, it's precisely there.

Looking at what you removed, I guess my question is: why would you use a blank line at all? What are the criteria? Why not suppress them altogether?

Regarding `type` attributes, [your link](https://stackoverflow.com/questions/6239032/do-you-really-need-to-specify-the-type-attribute/6239035#6239035) gives me reasons to keep them, not to remove them!
I'm sure you're not surprised&hellip; :¬D

> *&ldquo;HTML 4/XHTML [&hellip;] the type attribute is required.&rdquo;*
> *&ldquo;Browser support and server setups can't always be relied on - being explicit is a good idea as it avoids such problems.&rdquo;*
> *&ldquo;A simple solution for backwards compatibility is to add the attribute.&rdquo;*
> *&ldquo;Including it doesn't break validation, so you can convert to HTML5 from either HTML 4 or XHTML 1, and still have your `<script>` and `<style>` tags validate.&rdquo;*
> *&ldquo;Although browsers may be lenient, that's no reason to break convention.&rdquo;*

As I said before, basically for me

> *&ldquo;anything making the code/document clearer for the developer [and for the machines] is really only ever a good thing. If that means being explicit about the type of script being used or the type of style, I'd use it.&rdquo;*

I'm expecting you and @darobin to outvote me on this one again (and @darobin's vote alone beats rock, paper, scissors, Spock and lizard anyway).

--

Merging this nonetheless!
But OK to continue the discussion here for future reference etc.

astorije:
> What is the rationale for removing all those blank lines?

Man, you really love blank lines, don't you? :D

I would argue with the same arguments: I like clarity. Having blank lines from time to time to split sections is a good thing and makes things clearer, having a blank line at every other lines defeats the purpose.
Too many blank lines equals way too long source files, and therefore too much scroll. And scrolling among blank lines doesn't help reading the code, it's the opposite.
It's useful to have blank lines between `<article>` or `<section>` tags, for example, but not before closing `</body>` or `</html>` tags...
I have never seen many blank lines at the end of any other software and I doubt I will. However, a blank EOF is good practice and, as a matter of fact, many text editors will add one by default. That fulfills your command line argument. If you want more, it's the laptop owner's responsibility, by tweaking their `.bashrc` for example (and that is something you often see as a matter of fact). You cannot ask all the text files in the world to have extra blank lines at the end, can you? :-)

So you ask me a criteria: add blank lines for clarity, yes. But use too many of them and it just becomes noise and have the counter effect.

---

Regarding the `type` attribute, I should have said "read the facts and ignore the opinions" indeed :-)
Don't forget one thing: The answer was posted on June 2011, nearly 4 years ago, 3 years before HTML5 becomes a REC. And actually, at that time, [HTML5 was a WD](http://www.w3.org/TR/2011/WD-html5-20110525/)! Knowing that...

> *&ldquo;HTML 4/XHTML [&hellip;] the type attribute is required.&rdquo;*

HTML5 is neither HTML4n or XHTML so it cannot be compared.

> *&ldquo;Browser support and server setups can't always be relied on - being explicit is a good idea as it avoids such problems.&rdquo;*

It was a good idea when HTML5 was a WD. It is not anymore. If we still assume that a basic support of HTML5 cannot be relied on, then we should not trust *anything* HTML5 is adding...

> *&ldquo;A simple solution for backwards compatibility is to add the attribute.&rdquo;*

What backwards compatibility?

> *&ldquo;Including it doesn't break validation, so you can convert to HTML5 from either HTML 4 or XHTML 1, and still have your `<script>` and `<style>` tags validate.&rdquo;*

Sorry, that was supposed to be an argument in your favor? :D

> *&ldquo;Although browsers may be lenient, that's no reason to break convention.&rdquo;*

Well, not specifying this is now the convention ;-)

> *&ldquo;anything making the code/document clearer for the developer [and for the machines] is really only ever a good thing. If that means being explicit about the type of script being used or the type of style, I'd use it.&rdquo;*

I agree with the first sentence. The second one says the opposite to me. The code/document is made clearer when unnecessary information is removed. It leaves more room to actual, meaningful content.

Last thing: here we are talking about: 1) a tool that is and will always be used by the people building the web, presumably using bleeding-edge browsers, 2) an HTML file that will only be seen by those who make the mistake to go on the website rather than using the API 3) a test bed for development purposes... :-)

darobin:
So, I don't have a very strong opinion here, but just for the record.

I don't care about blank lines either way, it's pretty much a matter of personal taste. I tend to use some to separate large blocks, but I don't have a rule (I don't think I'm consistent myself).

For `type` I tend to prefer using it on `script` and `style` only when the content is something non-default (i.e. not JS or not CSS, typically a template, data, or a transpiled language). I find that makes it easier to spot the parts that are deliberately different. But it's not a strong feeling either, just a readability preference. I don't think that the arguments for keeping it in that SO post make much sense, even in HTML4 browsers already assumed the current defaults so I don't see anything ever breaking.

It's all just style really, the only thing that matters is consistency :)

PR #187, opened 2015-06-02
Fix path for TR installer

tripu:
Our script to install under /TR expects the second parameter to match `/^\/TR\/2015\//` (`2015` being the current year, of course).

But we didn't consider that in 7a59ece4e44637016daa3ac5f3844676f096ac65. Thus, [the error](https://lists.w3.org/Archives/Public/public-tr-notifications/2015Jun/0003.html).

This tries to fix that problem.


lib/orchestrator.js
```diff
@@ -16,6 +16,9 @@ var TokenChecker = require('./token-checker');
 // Configuration file
 require('../config.js');
 
+// Pseudo-constants:
+var W3_ORG_PREFIX = /^https?:\/\/(www\.)?w3c?\.org/i;
```
astorije: Ah ah, now that I am not around to grumble anymore, they are back! But, seriously, `http://w3c.org/`? And `http://WwW.w3C.OrG/`? Dude...

astorije:
This is not ideal, as the defaults now force the use of `/var/www`.
But yeah, if I introduced a bug in prod, it needs quick fixing indeed. I'll think of a more permanent solution after my current task on Echidna (and I cannot multitask now that I am "only" an open source contributor :-( ).

lib/orchestrator.js
tripu: I think I'm just applying the *robustness principle* here. You may argue that accepting `w3c.org` and `https` is stretching it too much, but I don't see good reasons not to be liberal with the input, as long as it doesn't pose a security risk. `/TR` will be HTTPS some day, and many users don't bother typing `www`. This regex acknowledges that. If all other pieces of the software followed this principle (flexible, case-insensitive regexes to filter inputs where that is reasonable) maybe we would have less issues like #190.

Issue #188, opened 2015-06-03
Wrong URL in e-mail notifications after successful publication


tripu:
The &ldquo;OK&rdquo; e-mail used to include the URL of the newly published spec ([example](https://lists.w3.org/Archives/Public/public-tr-notifications/2015Jun/0000.html)). After deploying [`v1.2.2`](https://github.com/w3c/echidna/releases/tag/v1.2.2) that template now includes the URL of the *source* used to publish instead ([example](https://lists.w3.org/Archives/Public/public-tr-notifications/2015Jun/0004.html)).

(Maybe a bug introduced during #84? I'll investigate.)

astorije:
Oops, most likely, sorry! :-)

PR #189, opened 2015-06-03
Fix wrong URL in e-mail notifications after successful publication

tripu:
See #188

### DIFF
--- a/lib/orchestrator.js
+++ b/lib/orchestrator.js
@@ -54,6 +54,30 @@ var Orchestrator = function (url, token, tempLoc, httpLoc, resultLoc) {
 };
 
 /**
+ * Pseudo-constant to indicate a step that finished OK.
+ */
+
+Orchestrator.STATUS_OK = 'ok';
+
+/**
+ * Pseudo-constant to indicate successful completion.
+ */
+
+Orchestrator.STATUS_SUCCESS = 'success';
+
+/**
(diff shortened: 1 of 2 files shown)


lib/orchestrator.js
```diff
@@ -54,6 +54,30 @@ var Orchestrator = function (url, token, tempLoc, httpLoc, resultLoc) {
...
+ */
+
+Orchestrator.STATUS_FAILURE = 'failure';
+
+/**
+ * Pseudo-constant to indicate errors.
+ */
+
+Orchestrator.STATUS_ERROR = 'error';
+
```
astorije: 4 lines in 23 lines... This is really frustrating...

tripu: What is it that you don't like here, @astorije?

Starting to substitute some string literals that appear several times in the code with static variables? The documentation comments for those variables? Or the blank lines in between?

astorije: Aaaaall this noise around the actual code :-)

```js
Orchestrator.STATUS_OK = 'ok';
Orchestrator.STATUS_SUCCESS = 'success';
Orchestrator.STATUS_FAILURE = 'failure';
Orchestrator.STATUS_ERROR = 'error';
```

is way enough considering the comments are just copy and paste of the keywords themselves :-)

PR #191, opened 2015-06-03
Add new dir for doc, and Remark presentation about Echidna

tripu:


astorije:
I don't think this should be stored in the code repository.

tripu:
I thought it was worth keeping a presentation about Echidna in the repo itself, as part of the documentation. We [do the same in Specberus](https://github.com/w3c/specberus/tree/master/doc). If it were stored under w3.org it would be more difficult for the team to collaborate on this. And having a separate repo just for that seemed too much to me.

What do you suggest, @astorije?

I have not problem moving it somewhere else.

astorije:
Valid points. I reckon this belongs to a `/Talks/` URL of the main website as it's, in the end, a talk about a tool.

One of the problems is that you will never keep this up-to-date (these are slides of a talk, not meant to change that much) so very quickly the document will be outdated.

If you put it in, say, `http://www.w3.org/Talks/`, it would be very appropriate to point to it in the index page of the wiki, or to have an extra page that lists links to all external URLs about Echidna.

In the end, just my 2c :-)

tripu:
OK, @astorije; I'll move it there after the talk, and link to it from the wiki as you suggest.

astorije:
Oki!
Similarly, please include remark.js and make it standalone instead of asking people to go to https://tripu.github.io/remark/remarkise?url=https%3A%2F%2Frawgit.com%2Fw3c%2Fechidna%2Fmaster%2Fdoc%2Fpresentation.md. In 2 years from now, I'd not be surprised if you get rid of your fork or if rawgit.com stops serving content.

tripu:
Come to think of it, I'll have to let the presentation stay here in the repo for a while&hellip; After sharing that URL I can't move it so soon.

PR #204, opened 2015-07-09
Check dependencies against nodesecurity.io DB as part of CI

tripu:
Run [`nsp audit-package`](https://github.com/nodesecurity/nsp#nsp-audit-package) during the CI cycle.

(This build fails only because of the reasons explained in #206.)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ before_script:
   - cp config.js.example config.js
 after_script:
   - npm run coveralls
+  - npm run nsp
 notifications:
   email: false
   irc:
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,9 +4,9 @@ node_js:
   - "0.10"
 before_script:
   - cp config.js.example config.js
+  - npm run nsp
 after_script:
   - npm run coveralls
-  - npm run nsp
 notifications:
   email: false
   irc:
--- a/package.json
+++ b/package.json
@@ -74,7 +74,7 @@
     "lint": "jshint .",
     "nsp": "nsp package",
     "start": "node app.js",
-    "test": "npm run check-style && npm run lint && mocha",
+    "test": "npm run nsp && npm run check-style && npm run lint && mocha",
     "testserver": "node test/lib/testserver"
   },
   "engines": {
--- a/package.json
+++ b/package.json
@@ -64,13 +64,15 @@
     "jscs": "^1.13.1",
     "jshint": "^2.7.0",
     "mocha": "^2.2.4",
-    "morgan": "^1.5.2"
+    "morgan": "^1.5.2",
+    "nsp": "^1.0"
   },
   "scripts": {
     "check-style": "jscs .",
     "coverage": "istanbul cover _mocha",
     "coveralls": "npm run coverage && cat ./coverage/lcov.info | coveralls",
     "lint": "jshint .",
+    "nsp": "nsp package",
     "start": "node app.js",
     "test": "npm run check-style && npm run lint && mocha",
     "testserver": "node test/lib/testserver"


tripu:
@deniak, this is ready to merge. The build is (rightly) broken; nsp is doing its job. I'm updating our dependencies in a separate PR to fix that in a moment.

(In spite of [what nsp recommends](https://github.com/nodesecurity/nsp#installation), I prefer to install nsp only locally. If nsp returns OK silently, it's because of [this issue in nsp](https://github.com/nodesecurity/nsp/issues/74#issuecomment-120215841). A solution may be to `rm -rf node_modules/nsp` and simply `npm install` again.)

astorije:
Ahoy! Quick stop here while I saw this PR appear in my notifications: note that this check is already done by [David](https://david-dm.org/w3c/echidna), which shows an `insecure` badge when something vulnerable is discovered.

It also wins over this addition in these 2 common use cases:

1. You do a commit on a Monday where no vulnerabilities have been discovered. One is discovered on the next Tuesday, but you don't get to commit anything for the next 2 weeks: no one knows about this vulnerability. In contrast, David's badge is updated as soon as a vulnerability is discovered, so anyone who ends up on the repository on GitHub (or npm, or wherever the README is showing up) can see it and open an issue. Granted that David does not notify by email, but it's still better than nothing.
2. The project `package.json` requires a package in version `1.0.0` or higher (like with a `^`), which has a well known vulnerability. Locally, you installed the package not long ago, when it was in version `1.0.1`. The bug was fixed so `nsp` doesn't complain. Same thing on Travis CI where the installed package is the most recent, say `1.0.2`. Now, on the production server, the package was already on version `1.0.0`, so running an update of npm there won't update anything. Et voilà, the vulnerability silently stays there.

Finally, since you added it in the `npm test` script, you'll probably grump at yourself in the future, same way I grumped at myself about adding the style check there. It reminds me of that express vulnerability about uploads (I think): we were not affected because there was no file upload mechanism, but changing version did require a tiny bit of work (probably just remove a deprecated call). Then this can happen while you are trying to commit and deploy something entirely different.

.travis.yml
```diff
@@ -4,6 +4,7 @@ node_js:
   - "0.10"
 before_script:
   - cp config.js.example config.js
+  - npm run nsp
```
astorije: This is run twice, as `npm run nsp` is also part of `npm test`.

On the other hand, the dependencies will **always** be up-to-date (against their major versions, which are never vulnerable) on the Travis CI instances, as it builds an environment from scratch using the latest packages.

astorije:
[Got bitten](https://github.com/w3c/echidna/pull/208#issuecomment-121414450), so :-1:, sorry to be that guy again :-)

tripu:
@astorije, thanks for the comments, and sorry again for having &ldquo;broken&rdquo; this with nsp&hellip;

David is a nice thing to have, and responds quicker to security issues, as you said. The problem with it is that probably nobody pays attention. Or maybe it's just me. I rarely stop by the home page, or check the README. Also, David now says [&ldquo;up to date&rdquo;, but with a yellow background](https://david-dm.org/w3c/specberus.svg). I guess that yellow colour is what indicates that [one package is outdated](https://david-dm.org/w3c/specberus)?&nbsp;&mdash;&nbsp;I hadn't noticed it before&hellip; Where is that &ldquo;insecure&rdquo; badge? I guess it's a discrepancy between the criteria David and nsp use to highlight issues. nsp digs down the whole tree, and thus even in Echidna it identifies the problem: Echidna > Specberus > &hellip; > &hellip; > qs. David doesn't see that problem.

As you say, packages will always be up-to-date in Travis. But that doesn't guarantee that there aren't vulnerabilities. We didn't notice that qs (supposedly) has problems until we added nsp to the build.

If I can fix that nsp error now (I asked for @deniak's help with that), I'd like to keep both systems in the CI cycle: one is quicker, the other one is more pedantic (I personally need the latter to act upon issues :) You may be right and we may find that this breaks builds too often, for spurious reasons. If that's the case, of course I'll remove nsp and see what can be done.

You're right about deprecated dependencies sometimes not really being an issue, but costing effort to update&hellip; We'll see about those case by case as usual, I guess.

Again, thanks for your comments, mate!

astorije:
@tripu, my main fear is that contributors will often not be able to run `npm test` and Travis CI builds will fail when they open a PR.
My point is that the toolset that runs when someone sends a PR should test against that PR. All other checks are noise that will scare people... So maybe keep it in the dependencies and script but remove it from `npm test`? :-/

Anyway, good luck with that! (I still have some code I wanted to submit, I can't wait to find the time to finish...)

tripu:
@astorije, why would contributors not be able to run `npm test`? Do you mean that `nsp` would fail and abort before getting to run our test suite with `mocha` if, eg, `nsp` detects issues (like is the case right now)?

If that's your concern, maybe we could customise the Travis CI script to have `nsp ` run only for PRs, and not for regular commits. The script would check the environment variable [`TRAVIS_PULL_REQUEST`](http://docs.travis-ci.com/user/pull-requests/#Security-Restrictions-when-testing-Pull-Requests) and build according to that. That way, all [correct] commits to a feature or bugfix branch will build OK (so contributors wouldn't be scared/confused), but the PR itself will fail if, eg, `nsp` is failing.

Well, I realise this is a bit convoluted&hellip; But now that I've learnt that builds can be customised in this way, I see how we could tweak the Travis config file to make things smoother :¬)

astorije:
My point is that if you make builds fail for something than what the PR is actually doing, you and the OP will never know if the PR actually passes. You'll have to fix the nsp issue somewhere else, merge it to master, and ask them to rebase on master to trigger a new build without the nsp issue.
In a nutshell: if a PR is good, its associated build should pass. Any failure based on something else than what the OP is submitting is noise.

Voilà, that was my conclusion on it. We'll see what history says.

PR #208, opened 2015-07-14
Remove unstable node version from the Travis CI builds

astorije:
#207 reminded me of this. I recently discovered that, for Node.js, even versions were stable and odd versions were unstable. So much for semantic versioning!

Currently, Node.js ecosystem is as follows:

- `0.12`: Stable
- `0.11`: Unstable, not maintained anymore
- `0.10`: Maintenance

See https://github.com/joyent/node/blob/master/ChangeLog.

Therefore, `0.11` can safely be removed from the build matrix.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: node_js
 node_js:
   - "0.12"
-  - "0.11"
   - "0.10"
 before_script:
   - cp config.js.example config.js


astorije:
Bloody `nsp`.

Well, `echidna@1.2.2 > specberus@1.1.1 > superagent@0.17.0 > qs@0.6.5`... Nothing I can do here (or should as part of this PR).

tripu:
Thanks, @astorije :¬)
(I'm sorry about the nsp annoyance&hellip; I switched it on assuming that I'd update packages easily to fix that error, but it's proving more difficult than that.)

PR #210, opened 2015-07-15
Upgrade all dependencies to latest version; exclude nsp from regular tests

tripu:
~~Except Specberus, which has to stay anchored to `1.1.0` until w3c/specberus#204 is fixed, and w3c/specberus#214 is merged.~~


tripu:
@deniak, I removed `nsp` from regular tests [as @astorije suggested](https://github.com/w3c/echidna/issues/212). We still run it in Travis to detect vulnerabilities

.travis.yml
```diff
@@ -4,6 +4,7 @@ node_js:
   - "0.10"
 before_script:
   - cp config.js.example config.js
+  - npm run nsp
```
astorije: I'd put it in the `after_script`, even after running the `coveralls` script, so that all tests and coverage reports are run even if there is a vulnerability that is not the PR author's fault and even if the coverage is bad and considered failing. And you still get your `nsp` report if test pass :-)

The only reason why `nsp` would not run is if the tests are failing, which is a valid cause to never merge and ask the author for changes anyway.

tripu: I thought of that, @astorije (doing it `after_script`). The reason I put it *before* is that I still think builds should break if `nsp` fails. Otherwise, we'd never see those scary red icons when `nsp` returned &ne; 0&hellip;

Vulnerabilities appear because of new security breaches being reported, or new insecure dependencies being introduced in the project. In the first case, the author of that PR simply has to restart the latest build for `master` (or for any other branch that was OK), which surely will fail too, to &ldquo;prove his/her innocence&rdquo;. In the other case, the build for that PR fails&hellip; and rightly so.

astorije: > I thought of that, @astorije (doing it `after_script`). The reason I put it *before* is that I still think builds should break if `nsp` fails. Otherwise, we'd never see those scary red icons when `nsp` returned &ne; 0&hellip;

Right, I though `after_script` would make the build fail too, my bad.

In that case, I would do:

```yaml
script:
  - npm test
  - npm run nsp
```

Which is actually a much better deal because both will be executed, even if one fails (I didn't know that, that's cool)! See the doc: http://docs.travis-ci.com/user/customizing-the-build/#Customizing-the-Build-Step

> Vulnerabilities appear because of new security breaches being reported, or new insecure dependencies being introduced in the project. In the first case, the author of that PR simply has to restart the latest build for `master` (or for any other branch that was OK), which surely will fail too, to &ldquo;prove his/her innocence&rdquo;. In the other case, the build for that PR fails&hellip; and rightly so.

That's a bit convoluted as a workflow IMO. Keep in mind that not everyone can restart a build (and even if they could, that's a weird thing to ask). In the present case (the command in `before_script`), fixing a dependency issue in master will require the PR author to rebase on it to see the tests passing. So you need to ask the author, wait for them to rebase, wait for the build to kick in, before being able to see the tests running. All of this, if the PR doesn't touch the `package.json` file, is not even necessary as the PR will not re-introduce the issue fixed in master. Anyway, you have my opinion, I'm just glad I am not merging anything anymore :D

tripu: Thanks, Jérémie. I'll put it *inside* `script` then soon, if @deniak is OK with that too. It seems like a good compromise.

Issue #213, opened 2015-07-22
Prepare code for new publication process (2015)


tripu:
cf [`http://www.w3.org/2015/04/Process-20150428/`](http://www.w3.org/2015/04/Process-20150428/)

tripu:
Fixed in #214.

PR #214, opened 2015-07-24
Adapt code to Process 2015

tripu:
As per [proposed process 1 September 2015](http://www.w3.org/2015/Process-20150901/).
See also [the roll-out plan](https://www.w3.org/Team/wiki/Process2015#Echidna_.26_Specberus).
This addresses #213.

This PR is intended as a place for tests and discussion for a while.
:warning: **Do not merge until 1 Sep 2015.**
:warning: **Will only work with Specberus `1.2.0` or greater**; remember to tag Specberus before merging and deploying this change in Echidna.

### DIFF
--- a/lib/specberus-wrapper.js
+++ b/lib/specberus-wrapper.js
@@ -47,7 +47,7 @@ SpecberusWrapper.validate = function (url) {
       validation: 'simple-validation',
       noRecTrack: false,
       informativeOnly: false,
-      processDocument: '2014'
+      processDocument: '2015'
     };
 
     if (process.env.NODE_ENV === 'dev') {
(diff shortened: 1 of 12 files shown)


tripu:
This PR expects Specberus to be at `1.2.0` at least; that tag is planned for 1 Sep 2015.

To check that this PR itself is OK, in spite of these errors, you can do this:

1. Edit [line #55 of `package.json`](https://github.com/w3c/echidna/blob/process2015/package.json#L55) to get the latest from `master` (instead of `1.2.0`, which does not exist yet). Write: `"specberus": "w3c/specberus"`.
2. `$ rm -rf node_modules/specberus/`
3. `$ npm install`
4. `$ npm test` again, and it'll pass OK.

tripu:
@deniak, after tagging Specberus `1.2.0` I launched the Travis builds again, and now they pass (as we expected).

Issue #135, opened 2015-03-19
UI status page


astorije:
This page is meant to debate about what should a status page look like.

Now fight!

tripu:
It should look awesome.

tripu:
The priority according to latest developments is to have a web front-end to the publication API, as an alternative (simpler) way for people to submit publication jobs. The focus is not so much on the status of all jobs right now.

cf  #297.

PR #227, opened 2015-09-08
Add method “stats” to the API and show usage stats on the main page

tripu:
This PR adds a new method to the API: `stats`, which accepts one optional parameter: `format`. It is useful to gather statistics about all the attempts to publish with the system so far. It works by finding all JSON files that correspond to past attempts to publish, and aggregating that information.

Those *attempts* are shown in two categories: successes, and failures. For now, stats are aggregated only per date (month by month) and per *deliverer*. An attempt with multiple deliverers counts toward the stats of all those deliverers. When info about deliverers is not available (because the JSON dump doesn't have the expected format, eg because it was produced by an old version of Echidna), this tool tries to infer it&nbsp;&mdash;&nbsp;by looking at the source URL, for example. Hyperlinks to those groups or sources are included when possible.

This also adds a new section to [the index page](https://labs.w3.org/echidna/); by clicking a link, the user can retrieve the latest stats and see them on that page.

Users will be able to open [`https://labs.w3.org/echidna/api/stats`](https://labs.w3.org/echidna/api/stats) too, to see those stats alone.

echidna-stats

Possible values of `format`:
* `html` (default): returns a simple HTML page with the stats.
* `div`: returns the stats within an element `<div id="echidna-stats">` (used to embed stats on the index page of Echidna).

This method, complemented with other methods or options to list all individual attempts, to search for specific attempts, and to gather detailed information about those, hopefully will replace the old need for a *dashboard*, *panel* or *status page* (see #135).

If this is merged, I'll upate the wiki to document this new feature.

**Questions:**
Are we disclosing too much by providing this method?
Shall the stats hide particular information about URLs and groups?

### DIFF
--- a/app.js
+++ b/app.js
@@ -51,7 +51,8 @@ app.get('/api/version', function (req, res) {
 app.get('/api/stats', function (req, res) {
   var format = (req && req.query && req.query.format) ? req.query.format : null;
   var stats = new Stats(format);
-  stats.run(function(data) {
+
+  stats.run(function (data) {
     res.send(data);
   });
 });
--- a/app.js
+++ b/app.js
@@ -20,6 +20,7 @@ var Job = require('./lib/job');
 var Orchestrator = require('./lib/orchestrator');
 var RequestState = require('./lib/request-state');
 var SpecberusWrapper = require('./lib/specberus-wrapper');
+var Stats = require('./lib/stats');
 
 // Configuration file
 require('./config.js');
@@ -45,6 +46,14 @@ app.get('/', function (request, response) {
 
 app.get('/api/version', function (req, res) {
   res.send(meta.version);
+});
+
+app.get('/api/stats', function (req, res) {
+  var format = (req && req.query && req.query.format) ? req.query.format : null;
+  var stats = new Stats(format);
+  stats.run(function(data) {
+    res.send(data);
+  });
 });
 
 app.get('/api/version-specberus', function (req, res) {
@@ -169,6 +178,8 @@ function corsHandler(req, res, next) {
   }
   next();
 }
+
+Stats.resultDir = argResultLocation;
 
 app.listen(process.env.PORT || port).on('error', function (err) {
   if (err) {
--- a/lib/stats.js
+++ b/lib/stats.js
@@ -1,19 +1,23 @@
 
 'use strict';
 
-var Fs   = require('fs');
+var Fs = require('fs');
 var Path = require('path');
 
 // Pseudo-constants:
-var FORMAT_HTML          = 'html';
-var FORMAT_DIV           = 'div';
-var DEFAULT_FORMAT       = FORMAT_HTML;
-var BOGUS_DOCUMENT_URL   = /^<documenturl>$/i;
-var UNKNOWN_VALUE        = '[n/a]';
+var FORMAT_HTML = 'html';
+var FORMAT_DIV = 'div';
+var DEFAULT_FORMAT = FORMAT_HTML;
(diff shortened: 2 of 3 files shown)


deniak:
Relying on the results stored on the filesystem can produce inconsistent results especially if multiple instances of echidna, not sharing the same mount point, are deployed.
I think it would be best to have the results stored in a database but that would also mean a major update.

tripu:
@deniak:
Definitely, having proper persistence is desirable.
In the meantime, `rsync`ing JSON result files between servers, or something like that, could do the trick.
Note that that is necessary, regardless of the outcome of this PR&nbsp;&mdash;&nbsp;eg to guarantee that calls to `api/status` will always give coherent results, regardless of which server is in charge in that particular moment&hellip;

deniak:
To be sure the same results are served, you would need to run that `rsync` pretty often to ensure consistency between the different servers.
Also, iterating through all the json files might lead to performance issues. If we have 10k+ json, and the endpoint is accessed by many people in a short period, this could be problematic.

On a side note, I'm a bit skeptical on why WGs/editors would need these stats. There's already a [tr-stats](http://www.w3.org/2002/01/tr-automation/tr-stats) form people can use.

tripu:
@deniak:

Wow, I didn't even know of [`w3.org/2002/01/tr-automation/tr-stats`](http://www.w3.org/2002/01/tr-automation/tr-stats)&hellip; That is handy. Although it doesn't say what came from Echidna, and what was manually published, right?

The motivations for these stats (in my mind) were two:
* For us developers, to keep track of recent activity, have a clear view of how many WGs/specs have adopted the system so far, detect spikes in the number of failures, plot the overall progress and increase in the number of users, etc.
* For editors and WGs, to have a sort of *Hall of Fame*, where they can ~~brag~~ show their activity, share with other groups to motivate them to get on board too, and quickly check if WG *x* or spec *y* are already being published with Echidna or not. I see it as part of the effort to get more groups using Echidna and spread the word.

But I agree that none of these goals are high priority.

Maybe @vivienlacourba or @tguild have suggestions about what to do here?

vivienlacourba:
**Disclaimer I know very little about echidna**

@tripu  I am like @deniak here.
I think people are more interested by stats of real publications (what we today have with [tr-stats](http://www.w3.org/2002/01/tr-automation/tr-stats) which could be exposed thru the W3C API) rather than stats of the tool used underneath.

This is like if we had on cvs.w3.org home page "hey look at pages that had the most commit" or "who has the most commits" (note that RRSAgent, steam and sysbot would likely trust the podium).

Can you just add a very simple list of groups using echidna on the homepage ?

Also if you guys were to implement those stats similar to @deniak's comment I would recommend having that info in a db rather than generated it based on local files (and even more if you need to use rsync).

@tripu from your last comment it seems there is a need for better info on echidna's use and failures. To me this is something I would expect to find in the application logs (meaning there might be enhancements to do on what echidna currently logs). If that is teh case you might want to discuss that in a separate github issue/

tripu:
Thanks, @vivienlacourba!

Then I'm (reluctantly) closing this PR. I intend to keep the branch alive though, as some of these changes might come handy in the future&nbsp;&mdash;&nbsp;when we tackle those other, more ambitious changes.

tguild:
I also think the stats are best in the SF publications system as exposed through existing tr-stats.  What may help with some of @tripu motivations is if we can record which were published through Echidna.  Adoption rate will be clearer since we have both Echidna and manual publications data in the same place.

Success/fail rate is also useful but not sure Echidna should send that upstream to publications system/tr-stats or store itself.

PR #229, opened 2015-09-18
Bump JSCS version and add new rule that the codebase already respects

astorije:
[Hey, it's me again.](https://www.youtube.com/watch?v=N0Ry1TwrM10)

(This will give you back a green badge, how supergreen is that!)


tripu:
Thank you, @astorije!

PR #238, opened 2015-11-19
Temporary fix for https://nodesecurity.io/advisories/46

tripu:
Merge this *after* w3c/specberus#268, to ensure (temporarily) that Echidna builds pass correctly.

To be undone once [the vulnerability](https://nodesecurity.io/advisories/46) has been fixed in all those other packages.


tripu:
Not needed any more.

PR #242, opened 2015-12-16
urls in manifest shouldn't be allowed

deniak:
If a manifest lists urls, echidna will create sub-folders:
```
index.html?specStatus=WD;shortName=foobar respec
https://www.w3.org/StyleSheets/TR/W3C-WD
https://www.w3.org/Icons/w3c_home
https://www.w3.org/StyleSheets/TR/logo-WD
```
see https://github.com/w3c/html-aria/pull/17/files

That PR forbids resources having `://` in its name.


plehegar:
:+1:

PR #241, opened 2015-12-08
Support for tarball

deniak:
That PR aims at allowing tarballs in echidna.
The tarball is expected to contain an `Overview.html` file as required by pubrules.

@tripu I'm assigning you as the reviewer but @plehegar, your feedback is welcome.

Note: `.tgz` and `.tar.gz` are not supported.

Fix https://github.com/w3c/echidna/issues/230

PR #246, opened 2015-12-21
support for tar upload

### LINKED ISSUE #243: Support direct uploading of the TAR file?
Apparently there was [a misunderstanding](https://lists.w3.org/Archives/Public/spec-prod/2015OctDec/0080.html) when we discussed [support for TAR in Echidna](https://github.com/w3c/echidna/issues/230): @tabatkins and others were expecting that TAR file to be passed directly to the API, while we in the W3C team interpreted that the TAR file would available for downloading somewhere :(

Shall we extend the API so that users can send their manifest or TAR file directly to it?

In principle, I'm all for it. I'm asking here to get people to raise early objections related to security, and to discuss the best way to add that to the API.

deniak:
PR #241 didn't include the tar upload to echidna. This is a [problem for some WGs](https://lists.w3.org/Archives/Public/spec-prod/2015OctDec/0080.html).
That PR should fix that.

To upload the tar via `curl`:
```
curl 'http://<echidnahost>/api/request' -F "tar=@/some/path/spec.tar" -F "decision=<decision>" -F "token=<token>"
```

fix #243

### DIFF
--- a/app.js
+++ b/app.js
@@ -37,7 +37,7 @@ app.use(bodyParser.urlencoded({ extended: false }));
 app.use(corsHandler);
 app.use(express.static('assets/'));
 
-// for parsing multipart/form-data
+// For parsing multipart/form-data
 var upload = multer();
 
 // Index Page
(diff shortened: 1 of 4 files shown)


tripu:
Thank you, @deniak!

It turns out I was working on my own implementation of this before @deniak submitted his PR (I assigned the issue to myself, but I should have been clearer communicating that, and commit my changes earlier&hellip;).

I'll see if there's anything useful from my own approach to add here. And also try to add tests for this (and at least keep code coverage at the level it was before).

tripu:
Merged as-is so that I don't block this any longer.

A couple of suggested improvements, for the record: #247, #248.

PR #255, opened 2016-02-25
Tar upload

deniak:
That MR add a new endpoint to handle tar upload. Tar upload requires basic auth.
That feature is currently locked as we need to find a way to easily extract the deliverers from the document being published and make sure the requester is one of the deliverer's participants.


lib/document-downloader.js
```diff
@@ -112,6 +112,9 @@ DocumentDownloader.fetchAndInstall = function (url, dest) {
           extract.end(content);
         });
       }
+      else if (type && type.mime === 'application/gzip') {
+        throw new TypeError('gzip\'ed tarballs not supported!');
```
tripu: I see these other common MIME types for compressed archives out there:
`application/x-gzip`, `application/zip`, `application/x-zip`, `application/bzip2`, `application/x-bzip2`.
Shall we check those, too?

tripu:
@deniak, I'm thinking if the auth part wouldn't be better encapsulated on a new separate module. To keep `app.js` short, and avoid mixing responsibilities&hellip;

lib/document-downloader.js
deniak: Good point. I fixed that part so now if a type is detected by `file-type` and is not 'application/x-tar', echidna will reject the request.
In case of a manifest and an html file, `file-type` will return a type `null`.

deniak:
@tripu I think the authentication part should be taken care of in express via the middleware. It's clear the endpoint requires authencation when you read the code.
However, when it comes to authorization, I think it should be dealt in a job in the orchestrator. As soon as the metadata extraction service is ready, I'll work on it. In the meantime, I think that MR can be merged.

tripu:
@deniak, I meant moving the passport setup to a new package, but no big deal.

Another thing: with this PR the API would expose these methods:

* `api/version`
* `api/version-specberus`
* `api/status`
* `api/request`
* `api/request/tar`

Which doesn't look consistent.

Can we keep only `api/request`, and decide whether to invoke `passport.authenticate()` or not by looking at some property of the request, or checking the value/presence of a new parameter? I'd rather add a parameter for this (eg, `type=tar`) than break the pattern we have for the methods of the API (`api/method?params`)&hellip;

tripu:
Maybe we can look at the `Content-Type` of the request, like [this](https://github.com/w3c/echidna/blob/tripu/support-upload/app.js#L89-L105)?

deniak:
@tripu good point. I just committed a fix and both methods should now use the same route.

tripu:
Thank you, @deniak.
I'm holding this one until I'm done with metadata extraction in Specberus, OK?

deniak:
That PR can be merged and deployed without breaking the current workflow but OK to wait for the other service.

PR #262, opened 2016-03-23
new jobs for tar upload and Notes publications

deniak:


tripu:
Ping me when it's ready for review, @deniak :)

deniak:
@tripu, I'm missing tests for the new jobs but you can already start the review.

tripu:
@deniak, with the current code, when I send wrong credentials on purpose, on the client I get

```
HTTP/1.1 500 Internal Server Error
X-Powered-By: Express
X-Content-Type-Options: nosniff
Content-Type: text/html; charset=utf-8
Content-Length: 942
Vary: Accept-Encoding
Date: Thu, 24 Mar 2016 10:05:22 GMT
Connection: keep-alive
```

and a stack trace; and the server spits

```
LDAP auth error: InvalidCredentialsError: Invalid Credentials
InvalidCredentialsError: Invalid Credentials
```

followed by another stack trace.

I think it's because even after you've caught the LDAP error, [you're still propagating it through the middleware to the `next` function (the param you call `done`)](https://github.com/w3c/echidna/blob/denis/tar-todos/app.js#L189).

If you replace that line with `done(null, user)`, the client receives

```
HTTP/1.1 401 Unauthorized
X-Powered-By: Express
WWW-Authenticate: Basic realm="Users"
```

and the server doesn't fail (just spits `LDAP auth error: InvalidCredentialsError: Invalid Credentials`).

deniak:
Good point @tripu !

Issue #265, opened 2016-04-12
improve results for tar upload


deniak:
Some parameters in the results are not available when uploading a tar. For e.g.:
https://lists.w3.org/Archives/Public/public-tr-notifications/2016Apr/0015.html

tripu:
(Closed via #268.)

PR #268, opened 2016-04-13
escape single quote in the result before passing it to the sendmail command

### LINKED ISSUE #267: JSON result (when error) is truncated
eg, [`lists.w3.org/Archives/Public/public-tr-notifications/2016Apr/0015.html`](https://lists.w3.org/Archives/Public/public-tr-notifications/2016Apr/0015.html).

The text stops *very close* to that apostrophe, but not quite there.

The full text that should have been sent is [here](https://lists.w3.org/Archives/Public/public-tr-notifications/2016Apr/att-0018/af7142d7-018a-4c6b-aa0b-2100973e838e.json).

deniak:
fix #267


deniak:
fix #265

PR #289, opened 2016-05-09
check if the memberOf regex matches before returning the group

deniak:
Some users belong to ldap groups that don't match the regex. Check if there's a match before returning the group.

### DIFF
--- a/lib/user-checker.js
+++ b/lib/user-checker.js
@@ -17,7 +17,9 @@ UserChecker.check = function (user, delivererIDs) {
   return new Promise(function (resolve) {
     var errors = new List();
     var groups = user.memberOf.map(function (group) {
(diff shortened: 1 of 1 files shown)


lib/user-checker.js
```diff
@@ -17,7 +17,8 @@ UserChecker.check = function (user, delivererIDs) {
   return new Promise(function (resolve) {
     var errors = new List();
     var groups = user.memberOf.map(function (group) {
-      return Number(group.match(/^cn=(\d*),ou=groups,dc=w3,dc=org$/i)[1]);
+      var matches = group.match(/^cn=(\d*),ou=groups,dc=w3,dc=org$/i);
+      if (matches) return Number(matches[1]);
```
tripu: Insert blank line between L20 and L21 to fix the CI (jscs) issue.

plehegar:
Yep, should fix the bug.

PR #291, opened 2016-05-23
sync metadata extracted by specberus and parameters passed to the backend

deniak:
Specberus now offers a service to extract all the metadata. The name of the parameters passed to the backend don't match what's expected. This PR fixes that.


tripu:
@deniak:

:+1:

&hellip;but please merge #295 on top of this one first
&hellip;which in turn depends on w3c/specberus#371.

deniak:
Tests will be fixed after we tag specberus

PR #321, opened 2016-06-21
fix(specberus-wrapper): return on rejection

marcoscaceres:
I think you meant to halt execution here, no?


deniak:
Thanks @marcoscaceres

PR #352, opened 2016-09-19
escape & in the result to avoid conflict with sendmail.sh

deniak:
Failure with results containing a `&` are not shown in the notification email because it's interpreted by bash. See https://lists.w3.org/Archives/Public/public-tr-notifications/2016Sep/0072.html


tripu:
I didn't know that `${//}` syntax&hellip; Thanks, @deniak!

PR #424, opened 2017-03-24
Implement --dry-run feature

### LINKED ISSUE #420: --dry-run feature
New optional parameter in the Echidna REST API: `--dry-run` (call it something else, aligned with the current syntax).

It will do two things:
* Make the process stop short of actual publication (ie, right before the `'publish'` step [here](https://github.com/w3c/echidna/blob/master/app.js#L108-L117). So, no real changes to `/TR` etc.
* Do *not* send any notification to the public ML. The result will be available as usual via the `status` method of the API. (And when we have custom mail addresses [as @dontcallmedom suggested](https://github.com/w3c/echidna/issues/308#issuecomment-223504778), the result will be sent there, too).

Intended to debug and test publication in a safe way.

tripu:
Closes #420.

The API understands a new optional parameter `dry-run`, and [interprets `true` if its value equals `'true'` (case-insensitive)](https://github.com/w3c/echidna/blob/tripu/dry-run/app.js#L88). Examples:

```bash
$ curl […] -d 'dry-run=true'
$ curl […] -d 'dry-run=TRUE'
$ curl […] -d 'dry-run=yes'    # false; equivalent to not setting the param at all
```

If `dry-run` is set to true, the document is not published, and no e-mail notification is sent (the only way to check results is then via the `status` method of the API).

If/when this is merged and deployed, I'll update [the wiki](https://github.com/w3c/echidna/wiki/How-to-use-Echidna) and I'll send a notice to the [spec-prod ML](https://lists.w3.org/Archives/Public/spec-prod/).

### DIFF
--- a/app.js
+++ b/app.js
@@ -144,26 +144,30 @@ var processRequest = function (req, res, isTar) {
       },
       requests[id].results
     ).then(function (state) {
-      var cmd = global.SENDMAIL + ' ' + state.get('status').toUpperCase() +
-        ' ' + global.MAILING_LIST;
-
-      if (state.get('status') === Orchestrator.STATUS_ERROR ||
-          state.get('status') === Orchestrator.STATUS_FAILURE) {
-        cmd += ' ' + (url || tar.originalname) + ' \'' +
-               JSON.stringify(requests[id], null, 2).replace(/'/g, '\\\'') +
-               '\'';
-      }
-      else {
-        cmd += ' ' + state.get('metadata').get('thisVersion') +
-          ' \'Echidna:   ' + meta.version +
-          '\nSpecberus: ' + SpecberusWrapper.version +
-          '\nJob ID:    ' + id +
-          '\nDecision:  ' + decision + '\'';
-      }
-
       console.log('[' + state.get('status').toUpperCase() + '] ' + url);
-      exec(cmd, function (err, _, stderr) { if (err) console.error(stderr); });
       dumpJobResult(argResultLocation + path.sep + id + '.json', requests[id]);
+      if (dryRun)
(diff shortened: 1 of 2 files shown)


app.js
```diff
@@ -103,18 +104,15 @@ var processRequest = function (req, res, isTar) {
...
-                 'third-party-checker', 'publish', 'tr-install',
-                 'update-tr-shortlink'];
-    }
+    if (isTar)
+      jobList.splice(2, 0, 'user-checker');
+    else
+      jobList.splice(3, 0, 'token-checker');
+
+    if (dryRun)
+      jobList.splice(jobList.indexOf('publish'));
```
deniak: also remove the jobs 'tr-install' and 'update-tr-shortlink'

app.js
```diff
@@ -85,6 +85,7 @@ var processRequest = function (req, res, isTar) {
   var token = (!isTar && req.body) ? req.body.token : null;
   var tar = (isTar) ? req.file : null;
   var user = req.user ? req.user : null;
+  var dryRun = Boolean(req.body && req.body['dry-run'] && /^(true|t|yes|y|1|on)$/i.test(req.body['dry-run']));
```
deniak: I'm not convinced allowing multiple value for dry-run is really useful. I would go with `true` only but no big deal.

app.js
tripu: @deniak, it does that ([`Array.prototype.splice()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/splice#Parameters)).

deniak: Oh right!  :-/

PR #430, opened 2017-04-18
Group participation fix

### LINKED ISSUE #429: Echidna wrongfully requires membership of all groups in joint publication
As illustrated in [this attempt to publish a Media Capture Task Force document](https://lists.w3.org/Archives/Public/public-tr-notifications/2017Apr/0013.html) (a joint task force between DAS and WebRTC), echdina currently requires membership of both groups in joint publications, whereas I can't think of any good reason to impose this.

deniak:
Echidna should check that the requester is participating in *at least* one WG delivering the spec.
Fix #429 
cc @dontcallmedom

### DIFF
--- a/lib/user-checker.js
+++ b/lib/user-checker.js
@@ -23,9 +23,7 @@ UserChecker.check = function (user, delivererIDs) {
     }));
     var delivererSet = new Set(delivererIDs);
 
-    var participation = [...groups].filter(x => delivererSet.has(x));
-
-    if (participation.length === 0) {
+    if (![...groups].some(x => delivererSet.has(x))) {
       errors = errors.push('You are not participating in the following groups: ' +
           delivererIDs.join(', '));
     }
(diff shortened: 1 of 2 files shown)


lib/user-checker.js
```diff
@@ -16,17 +16,18 @@ var UserChecker = {};
...
+    var delivererSet = new Set(delivererIDs);
 
-    for (var i = 0; i < delivererIDs.length; i++) {
-      if (groups.indexOf(delivererIDs[i]) < 0) {
-        errors = errors.push('You are not participating in group: ' +
-          delivererIDs[i]);
-      }
+    var participation = [...groups].filter(x => delivererSet.has(x));
+
+    if (participation.length === 0) {
```
tripu: These two lines and the variable could be replaced by this one:

```javascript
if ([...groups].some(x => delivererSet.has(x))) {
```

deniak: Thanks for the suggestion. I updated the PR.

Issue #231, opened 2015-09-23
Readable error messages


deniak:
From [Bert's whishlist](https://lists.w3.org/Archives/Public/public-pubrules-comments/2015Sep/0000.html)

> Readable error messages, preferably a format close to what compilers 
> generate, so you can use them inside Emacs or some other IDE.

PR #431, opened 2017-04-21
Improve mail notifications

### LINKED ISSUE #308: Send better e-mail, and simplify how we do it
- [ ] **More descriptive and succinct**: explain upfront what was published; or else which step failed, and how exactly it failed. Don't dump the JSON result.
- [ ] **Attach the complete JSON report** for details and debugging.
- [ ] **Format mail** for better readability (either HTML+CSS, or at least compose it better in plain text).
- [ ] **Get rid of [the fragile Bash script](https://github.com/w3c/echidna/blob/master/sendmail.sh)** and instead send mail directly from Node.js. Some packages to explore:
  - [`nodemailer`](https://www.npmjs.com/package/nodemailer).
  - [`pixl-mail`](https://www.npmjs.com/package/pixl-mail).
  - [`x-mail`](https://www.npmjs.com/package/x-mail).
  - [`staticcss`](https://www.npmjs.com/package/staticcss).
- [x] **Include the URL of the &ldquo;decision to publish&rdquo;** as suggested by @aaronpk (w3c/specberus#529).

tripu:
# Summary

* Gets rid of Bash script.
* Uses [`nodemailer`](https://www.npmjs.com/package/nodemailer) and [`nodemailer-sendmail-transport`](https://www.npmjs.com/package/nodemailer-sendmail-transport) instead (can use either local `sendmail` or an SMTP server).
* Plain-text *and* HTML versions of messages via [Foundation for Emails 2](http://foundation.zurb.com/emails.html); messages [inlined](http://foundation.zurb.com/emails/inliner-v2.html).
* New configurable settings:
  * Plain text only vs. also HTML version.
  * Attached JSON yes/no (rationale: see below).
  * Address to use in `from:` field.
* Errors are more readable.
* Includes in message:
  * URL to corresponding `status` API method on server (so that even if the JSON is *not* attached, details are available to the user anyway).
  * (If failure) links to specific Specberus rules that did not pass (when possible).
  * (If failure) link to a new pre-filled issue to submit, to report bugs ([example](https://github.com/w3c/echidna/issues/new?title=Bug:%20%5BWHAT%5D&body=%5BEXPLANATION%5D%0A%0AFound%20while%20trying%20to%20publish%20%60https%3A%2F%2Fwww.w3.org%2FTR%2Fpayment-request%2F%60%20(see%20the%20%5Bpublication%20job%5D(https%3A%2F%2Flabs.w3.org%2Fechidna%2Fapi%2Fstatus%3Fid%3Dad5059e5-0f3b-4a92-9cff-e5ec54a4a997)).&labels=from-template)).

# Examples of sent messages

## Failure, plain text

Subject: `[Echidna] ✗ Failure: https://foo.bar/specs/latest`

```
✗ Echidna failed to publish https://foo.bar/specs/latest
  for the following reasons:

    The file has been retrieved.
    The metadata have been extracted.
    The document failed Specberus.
  👎 Error code “wrong-date”.
      ⮡ cf https://www.w3.org/pubrules/doc/rules/?profile=WD#dateTitleH2
  👎 HTTP resource not retrievable. The HTTP status from the remote server was: 404.
      ⮡ cf https://www.w3.org/pubrules/doc/rules/?profile=WD#valideHTML

More details are available here:

  https://labs.w3.org/echidna/api/status?id=ad5059e5-0f3b-4a92-9cff-e5ec54a4a997

If you think you found a bug, please fill in and submit this issue:

  https://github.com/w3c/echidna/issues/new?title=Bug:%20%5BWHAT%5D&body=%5BEXPLANATION%5D%0A%0AFound%20while%20trying%20to%20publish%20%60https%3A%2F%2Fwww.w3.org%2FTR%2Fpayment-request%2F%60%20(see%20the%20%5Bpublication%20job%5D(https%3A%2F%2Flabs.w3.org%2Fechidna%2Fapi%2Fstatus%3Fid%3Dad5059e5-0f3b-4a92-9cff-e5ec54a4a997)).&labels=from-template

~ Fri Apr 21 2017 18:15:12 GMT+0200 (CEST)

-- 
Echidna 1.11.10 (Specberus 3.2.5)
Copyright © 2014–2017 W3C — https://www.w3.org/
```

## Success, plain text

Subject: `[Echidna] ✔️ Success: https://www.w3.org/TR/payment-request/`

```
✔️ Echidna just published “Payment Request API”
  (payment-request), a WD:

  👍 https://www.w3.org/TR/payment-request/

More details are available here:

  https://labs.w3.org/echidna/api/status?id=be457ba0-1053-49c3-94a5-6c6e15d2a021

~ Fri Apr 21 2017 18:16:41 GMT+0200 (CEST)

-- 
Echidna 1.11.10 (Specberus 3.2.5)
Copyright © 2014–2017 W3C — https://www.w3.org/
```

Fixes #308.


tripu:
Missing the &ldquo;decision to publish&rdquo; now; I'll add it.

tripu:
Fixes #474.

tripu:
@deniak, I see a lot of value in sending HTML instead of plain text. I'm not thinking of making it look prettier: I think reports will be much easier to scan quickly, and to understand in detail, when there are colours, typography, a good layout, etc. Much easier for the user to answer their own questions: &ldquo;was my publication successful?&rdquo; &ldquo;if not, how many errors are there?&rdquo; &ldquo;how important are they?&rdquo; &ldquo;are there any links here to help me diagnose the problem?&rdquo; &mdash; etc.

It's like going from this (unrelated example):

---

Example 1

---

to this (unrelated example):

---

Example 2

---

The latter is much more readable.

That said, I know I have lost all credibility on this subject, given how much I have neglected this PR&hellip;

So, :+1: for now to forget about HTML and to rework the plain-text version only. I hope I can pick up my changes here in the future, to submit a separate PR that will be specific about HTML e-mail!

deniak:
@tripu I agree HTML can help emphasize errors or specific sections but I'm just worried about the maintenance, compatibility and a11y compliance.
If we can make sure it all works fine, then I'm ok switching to HTML and drop plain text.

tripu:
Just to clarify: I never meant to drop the plain-text version.

Messages would always be sent with both `Content-Type`s: `text/plain` and `text/html`. That should help to keep messages always compatibility with old MUAs, and with A11Y.

But yes: that introduces a bit of extra complexity, and we'd have to maintain 2&times; as many message &ldquo;templates&rdquo;.

deniak:
AFAIK, there can be [only one content type](https://tools.ietf.org/html/rfc2616#section-14.17).

:-1: to supporting both html and plain text. I can't think of a good use case where it will be useful. I don't think that project will be used outside W3C as is, so there's no need to provide support for both formats.

tripu:
@deniak, I mean `multipart` mail like this:

```
Date: Sun, 16 Sep 2018 10:50:13 +0000 (UTC)
From: Foo <foo@foo.fooz>
To: Bar <bar@bar.baz>
Message-ID: <13048957238945720495824572854>
Subject: Foo bar
Mime-Version: 1.0
Content-Type: multipart/alternative;
 boundary="--==_mimepart_398475827894753";
 charset=UTF-8
Content-Transfer-Encoding: 7bit


----==_mimepart_398475827894753
Content-Type: text/plain;
 charset=UTF-8
Content-Transfer-Encoding: quoted-printable

This is *great news*!
=0D
----==_mimepart_398475827894753
Content-Type: text/html;
 charset=UTF-8
Content-Transfer-Encoding: quoted-printable

<!DOCTYPE html>
  <html>
    <p>
      This is <em>great news</em>!
    </p>
    <img src=3D"https://example.com/foo.png" alt=3D"An image" />
  </html>
=0D
----==_mimepart_398475827894753--
```

This keeps both humans and machines happy. The text version is useful for users with reading disabilities or visual impairments; for text-based MUA's, narrow screens, low-end devices and poor network connections; and for scripts or hooks written to process these notifications automatically&nbsp;&mdash;&nbsp;while the rich text version will be shown in all other cases.

I think [Nodemailer supports this very easily](https://nodemailer.com/message/).

vivienlacourba:
Assigning to @deniak as @tripu is focusing on other projects (helix).

tripu:
> *&ldquo;Also that PR is supposed to fix #231 so this use case should be the focus.&rdquo;*

The focus of this PR was on the ideas in #308. Those are improvements requested by several users over time.

#231 is something that only @bert-github wants, afaik.

> *&ldquo;Reviewing this code, the HTML email part is clearly a WIP (eg `nodemailer` `message.html` expect the html body content not a full html page, or HTML emails should use inline css rules for being properly rendered on the various mailer).&rdquo;*

I'm not sure I understand what you mean. Although I did not finish all 4 templates (*{text, html}&nbsp;&times;&nbsp;{success, failure}*), the mechanism is in place and in my tests seemed to work well. The templates are complete HTML pages, including embedded Foundation CSS: [failure](https://github.com/w3c/echidna/blob/1aa473d7c07bea4b17b91e03cc823dac269b706d/assets/mail-templates/failure.html); [success](https://github.com/w3c/echidna/blob/1aa473d7c07bea4b17b91e03cc823dac269b706d/assets/mail-templates/success.html).

> *&ldquo;Side note it seems the current code uses manually defined templates, using a templating engine to generate the body (be it text or html) seems the proper the way to go (eg to have a least a common footer).&rdquo;*

I'm not using a template engine because I found that for our needs here [3 LoC were enough](https://github.com/w3c/echidna/blob/1aa473d7c07bea4b17b91e03cc823dac269b706d/lib/mailer.js#L89-L91) to interpolate the templates.

The only manual step I wanted to improve here is that right now I'm generating the final templates [using this online tool by Foundation](https://foundation.zurb.com/emails/inliner-v2.html). That, I should do programmatically somehow&hellip;

> *&ldquo;So I would be in favor of dropping it for now if it takes too much time to @deniak . HTML email can have some benefits here so it could be tackle as a future enhancement in a separate PR.

:+1:

deniak:
I have fully tested that PR but my understanding is that all the pieces are there but it might require more testing to make sure the HTML emails are displayed ok in the different clients. @vivienlacourba also noticed the [`nodemailer` documentation](https://nodemailer.com/message/) talks about the HTML body only and not the full HTML page.

Regarding the template engine, I would :+1: using one, as it seems a cleaner solution and it can probably be helpful to merge the text and html templates. Since we recommend, handlerbars, there's a module for nodemailer [available](https://www.npmjs.com/package/nodemailer-express-handlebars).

Anyhow, I'll move the code related to HTML emails in a separate PR for now.

tripu:
> *&ldquo;@vivienlacourba also noticed the [`nodemailer` documentation](https://nodemailer.com/message/) talks about the HTML body only and not the full HTML page.
Regarding the template engine, I would :+1: using one, as it seems a cleaner solution and it can probably be helpful to merge the text and html templates. Since we recommend, handlerbars, there's a module for nodemailer [available](https://www.npmjs.com/package/nodemailer-express-handlebars).&rdquo;*

After I talked with @vivienlacourba yesterday, I came to understand these two points (passing the `body` to `nodemailer` only; and the benefits of a template engine). So :+1: to both; and yes, I'll use [`nodemailer-express-handlebars`](https://www.npmjs.com/package/nodemailer-express-handlebars) if/when I submit a new PR with the HTML extension.

deniak:
@tripu ok, I'll add a new commit to integrate handlebars

assets/mail-templates/failure.txt
```diff
@@ -0,0 +1,17 @@
...
+
+If you think you found a bug, please fill in and submit this issue:
+
+  https://github.com/w3c/echidna/issues/new?title=Bug:%20%5BWHAT%5D&body=%5BEXPLANATION%5D%0A%0AFound%20while%20trying%20to%20publish%20@ENCODEDSOURCE%20(see%20the%20%5Bpublication%20job%5D(https%3A%2F%2Flabs.w3.org%2Fechidna%2Fapi%2Fstatus%3Fid%3D@ID)).&labels=from-template
+
+~ @TIMESTAMP
+
+-- 
+Echidna @ECHIDNA (Specberus @SPECBERUS)
+Copyright © 2014–2017 W3C — https://www.w3.org/
```
tripu: s/2017/2018/

assets/mail-templates/success.txt
```diff
@@ -0,0 +1,14 @@
...
+
+More details are available here@ATTACHMENT:
+
+  https://labs.w3.org/echidna/api/status?id=@ID
+
+~ @TIMESTAMP
+
+-- 
+Echidna @ECHIDNA (Specberus @SPECBERUS)
+Copyright © 2014–2017 W3C — https://www.w3.org/
```
tripu: s/2017/2018/

assets/mail-templates/failure.txt
```diff
@@ -0,0 +1,17 @@
+✗ Echidna failed to publish @SOURCE
+  for the following reasons:
+
+@ERRORS
+More details are available here@ATTACHMENT:
+
+  https://labs.w3.org/echidna/api/status?id=@ID
```
tripu: I wonder if we should use `global.BASE_URI` or `process.env.BASE_URI` here, instead of hard-coding the URL?

assets/mail-templates/success.txt
```diff
@@ -0,0 +1,14 @@
+✔️ Echidna just published “@TITLE”
+  (@SHORTNAME), a @TYPE:
+
+  👍 @URL
+
+More details are available here@ATTACHMENT:
+
+  https://labs.w3.org/echidna/api/status?id=@ID
```
tripu: I wonder if we should use `global.BASE_URI` or `process.env.BASE_URI` here, instead of hard-coding the URL?

app.js
```diff
@@ -151,8 +152,8 @@ var processRequest = function (req, res, isTar) {
       Orchestrator.hasFinished,
       function (state) {
         requests[id].results = state;
-        console.log(JSON.parse(JSON.stringify(requests[id])));
-        console.log('----------');
+        // console.log(JSON.parse(JSON.stringify(requests[id])));
+        // console.log('----------');
```
tripu: These are remains of my test&hellip; I think the output in the console is too verbose.
Shouldn't we remove a few `console.log()` now?

assets/mail-templates/failure.txt
deniak: I think that's ok to keep the URL in the template unless we think it'll be used in many different places.

app.js
deniak: The `console.log` are commented out. You prefer to remove them completely?

assets/mail-templates/failure.txt
```diff
@@ -12,6 +12,6 @@ If you think you found a bug, please fill in and submit this issue:
 
 ~ @TIMESTAMP
 
--- 
+--
```
tripu: I think `--` (plus a trailing space) is kind of a standard ([link](https://en.wikipedia.org/wiki/Signature_block#Signatures_in_Usenet_postings), [link](https://together.jolla.com/question/2884/email-signature-separator-use-standard-dash-dash-space-/)).
No sure if some MUA out there will *not* parse this signature correctly if it doesn't have that space&hellip;?
(Same comment about the other template.)

deniak: ah, sorry, my editor is configured to remove the trailing whitespaces automatically.

app.js
deniak: Done @tripu

PR #467, opened 2017-07-14
Simple Web interface

### LINKED ISSUE #297: form to submit a request to Echidna
People who aren't familiar with CLI could benefit from a simple form to submit their documents to Echidna. How does it sound?

tripu:
I know this is very low priority, but the work was done already. And because I just fixed a small, long-outstanding bug in this web UI, I thought perhaps we could merge this into `master`?

Most changes are old; have been in this branch for ages. I just committed a couple improvements/fixes, and squashed it all together for simplicity.

`$ npm start`, then open `http://localhost:3000/ui`.

While at it, update jQuery from 2.1.4 to 2.2.4 (it's used by `views/index.html` and by the abandoned `test/views/index.html`, too).

Closes #297.


deniak:
@tripu I want to test the interface thoroughly but here are also a few remarks:
- shouldn't we use the jquery at https://www.w3.org/scripts/?
- the UI doesn't support the tar upload method. Note that I'm fine leaving that feature for later

tripu:
@deniak:

> *&ldquo;Shouldn't we use the jquery at https://www.w3.org/scripts/?&rdquo;*

You're right. The same about that Bootstrap CSS. I'll change that. Wait.

> *&ldquo;The UI doesn't support the tar upload method. Note that I'm fine leaving that feature for later.&rdquo;*

OK, let's leave for a later PR. As agreed, there isn't much demand for this web UI, anyway.

deniak:
@tripu 
>> “Shouldn't we use the jquery at https://www.w3.org/scripts/?”

> You're right. The same about that Bootstrap CSS. I'll change that. Wait.

I see that PR still use a local bootstrap and jquery

tripu:
@deniak, I've made the modifications we discussed.

PR #507, opened 2017-10-13
chore(package): update mocha to version 4.0.1

tripu:
Closes #503


tripu:
Build errors seem to come from this version of Mocha; cf mochajs/mocha#3067

Does not seem specific to Travis; when I test locally: `npm test` does not return; it stays frozen.

tripu:
Fails both on Travis and on my local tests. I have tried a couple things, but I don't have the time to try to debug properly. Perhaps mochajs/mocha#3067, or related to it.

PR #508, opened 2017-10-13
fix(package): update moment to version 2.19.1

tripu:
Closes #504


tripu:
[Tested automatically](https://coveralls.io/builds/13706428/source?filename=lib%2Fpublisher.js#L35)

PR #509, opened 2017-10-13
fix(package): update file-type to version 7.0.1

tripu:
Closes #427


tripu:
[Tested automatically](https://coveralls.io/builds/13411986/source?filename=lib%2Fdocument-downloader.js#L82)

PR #531, opened 2017-12-29
fix(package): update moment to version 2.20.1

tripu:
Closes #527


tripu:
[Tested automatically](https://coveralls.io/builds/14836736/source?filename=lib%2Fpublisher.js#L41)

PR #534, opened 2018-01-22
fix(package): update file-type to version 7.5.0

tripu:
Closes #427


tripu:
(`file-type` is used in the test suite.)

PR #542, opened 2018-02-19
chore(package): update nsp to version 3.2.1

tripu:
Closes #539


tripu:
Tested on Node.js 8, and :+1:
[The new nsp vulnerability](https://nodesecurity.io/advisories/566) seems to affect only versions of package `hoek` that get installed on Node.js 6.

PR #547, opened 2018-03-23
Fix tests after switch to process 2018

deniak:
Two tests are failing because we no longer accept documents on process 2017. That PR switches these test documents to process 2018.


deniak:
The NSP error will be fixed when https://github.com/w3c/third-party-resources-checker/pull/25 is merged and we submit a new version of the third-party-checker

PR #548, opened 2018-04-03
fix(package): update moment to version 2.22.0

deniak:
Closes #543


tripu:
Tested automatically in `lib/publisher.js`.

PR #549, opened 2018-04-05
upgrade third party checker

deniak:


deniak:
@tripu good point. Done!

PR #551, opened 2018-04-09
fix(package): update request to version 2.85.0

tripu:
Closes #545


tripu:
Tested automatically in `/lib/document-downloader.js` and `/lib/token-checker.js`.

PR #560, opened 2018-04-30
fix(package): update tar-stream to version 1.6.0

tripu:
Closes #558


tripu:
[Covered by automatic tests](https://coveralls.io/builds/16732019/source?filename=lib%2Fdocument-downloader.js#L90)

PR #565, opened 2018-05-18
Update to node 10 in .travis.yml

tripu:


package.json
```diff
@@ -45,9 +45,9 @@
   "dependencies": {
     "body-parser": "1.18.2",
     "compression": "1.7.2",
-    "express": "5.0.0-alpha.6",
+    "express": "4.16.3",
```
deniak: Why using an older version?

package.json
```diff
@@ -45,9 +45,9 @@
...
     "file-type": "7.7.1",
-    "immutable": "4.0.0-rc.4",
+    "immutable": "3.8.2",
```
deniak: Why using an older version?

package.json
tripu: https://github.com/w3c/specberus/pull/752#discussion_r189235375

package.json
tripu: https://github.com/w3c/specberus/pull/752#discussion_r189235375

PR #567, opened 2018-05-24
fix(package): update tar-stream to version 1.6.1

tripu:
Closes #558


tripu:
Branch rebased and OK now.
`tar-stream` [tested automatically](https://coveralls.io/builds/17141297/source?filename=lib%2Fdocument-downloader.js#L90).

PR #568, opened 2018-05-24
fix(package): update file-type to version 8.0.0

tripu:
Closes #427


tripu:
[Tested automatically](https://coveralls.io/builds/17141397/source?filename=lib%2Fdocument-downloader.js#L82)

PR #569, opened 2018-05-24
fix(package): update request to version 2.87.0

tripu:
Closes #563


tripu:
Tested automatically ([`lib/token-checker.js#L25`](https://coveralls.io/builds/17141833/source?filename=lib%2Ftoken-checker.js#L25), [`lib/document-downloader.js#L18`](https://coveralls.io/builds/17141833/source?filename=lib%2Fdocument-downloader.js#L18)).

PR #571, opened 2018-05-29
Update chai-immutable to v2

astorije:
Thanks for letting your CI be my guinea pig while I test the v2 :D
(I suggest you do not merge until v2, but I want to see a green check on your end before being confident to release v2)

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,7 +64,7 @@
   "devDependencies": {
     "chai": "4.1.2",
     "chai-as-promised": "7.1.1",
-    "chai-immutable": "1.6.0",
+    "chai-immutable": "2.0.0-rc.2",
     "coveralls": "3.0.1",
     "istanbul": "0.4.5",
     "jscs": "3.0.7",


tripu:
@astorije, hi! :)

My updated `master` builds locally all right, and [on Travis CI, too](https://travis-ci.org/w3c/echidna/builds/383151061) (I just restarted it now). Are you sure you wiped your `node_modules/` and `npm i`'d after you checked out `master` locally, and before testing&hellip;?

astorije:
@tripu, I'm not sure, but that wouldn't change Travis' report :/

tripu:
@astorije, I'm still a bit confused, but this is what I saw today:

Now, when I check out your branch, remove `node_modules` and `package-lock.json`, install dependencies from scratch, and run the tests, I get only one error, and one that is different from [the errors Travis reports](https://travis-ci.org/w3c/echidna/builds/385045451):

    124 passing (11s)
    1 failing

    1) DocumentDownloader
         fetch(url)
           should reject if the server is not reachable:
       Error: Timeout of 10000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (echidna/test/test.js)

Simply by giving Mocha an extra second (like in https://github.com/w3c/echidna/pull/573/files#diff-9c115eb1986d9657744029cc6ca59c89), that error goes away.

Try that?

astorije:
This is really weird. I removed my `node_modules` and `package-json.lock` to no effect, still get this locally:

```
  1) SpecberusWrapper
       validate(url-with-css-errors)
         should return an error property that has 4 errors:

      AssertionError: expected { Object (size, _origin, ...) } to have size 4 but got 5
      + expected - actual

      -5
      +4



  2) SpecberusWrapper
       validate(url-with-css-warnings)
         should return an error property that has 2 error:

      AssertionError: expected { Object (size, _origin, ...) } to have size 2 but got 3
      + expected - actual

      -3
      +2
```

To be noted that I do not have Pubrules or whatnot, and therefore set `API_KEY=foo`. I can also confirm that the same steps on the latest `master` give the exact same result locally. I am really puzzled and pretty confident this is not coming from my PR :/

Oh and I confirm that I don't see a difference when increasing Mocha's timeout!

astorije:
Sidenote: `Object (size, _origin, ...) }` is really not helpful. If only the idiot who maintains `chai-immutable` could improve this! :)

tripu:
> To be noted that I do not have Pubrules or whatnot, and therefore set API_KEY=foo

I can reproduce your two errors if I set a bogus API key. With a valid key, `tripu/fix-dependencies` builds all right (and Travis CI builds `master` OK with a valid key, too). It must be that.

Man, [get a key](https://www.w3.org/users/myprofile/apikeys), it's free! :)

astorije:
Oooooh, you set https://github.com/w3c/echidna/blob/master/.travis.yml#L8 on the Travis config side 😱
So in short there is no way this PR can fully pass :S

Can you test my PR with your valid API key?

For Travis CI config, I think you should be able to write `API_KEY` in the Travis config file, and encrypt it with https://docs.travis-ci.com/user/encryption-keys/ so that PRs would pass too.

tripu:
@astorije: yes, if I check out your branch, **apply [my fix for `mocha.opts`](https://github.com/w3c/echidna/pull/573/files#diff-9c115eb1986d9657744029cc6ca59c89)**, and run the tests with my valid API key, it succeeds.

I'm merging your PR (it should succeed once it's in `origin/master`).

And I'll see how to fix that encrypted key, yes. Thanks for pointing that out!

PR #575, opened 2018-06-14
CR: check CR with specberus and check approvals with GH API

### LINKED ISSUE #522: Support CR profile
High-level meta-issue to keep track of tasks involved:

* [ ] #523 (automate approval of CR requests by the project manager):
Echidna API / Echidna web UI / **GH issues** ⬅️ / some other approach.
* [ ] How to send the CfE:
**Do from new IPP Symfony code** ⬅️ / do in script / do from Echidna / &hellip;
* [ ] #524 Give submitters a mechanism to indicate whether changes are purely editorial, or substantive:
There will be a new parameter for that.
* [ ] w3c/specberus#39 (improve detection of preliminary implementation/interoperability reports).
* [ ] Detect if it's a CR coming from an edited REC (in such case the *shortlink* does not change; see diagram below).

diagram

### LINKED ISSUE #524: Add parameter to API to indicate substantive or editorial change

deniak:
Fix #522 (workflow described in the tickets)
Fix #524 

This PR adds support to publish CRs in echidna.
There's an additional optional parameters to specify if the CR contains substantive changes or not. By default, we assume it does.


lib/publisher.js
```diff
@@ -46,7 +45,11 @@ Publisher.prototype.publish = function (metadata) {
       rectrack: metadata.get('rectrack'),
       informative: metadata.get('informative'),
       editorDraft: metadata.get('editorDraft'),
-      processRules: metadata.get('processRules')
+      processRules: metadata.get('processRules'),
+      implementationReport: metadata.get('implementationReport'),
+      implementationFeedbackDue: new Moment(metadata.get('implementationFeedbackDue', 'YYYY-MM-DD')
+              .format('YYYY-MM-DD'),
```
tripu: There's a missing `)` in this expression, and I think the first param `'YYYY-MM-DD'` shouldn't be there?

Apart from that, why
`new Moment(metadata.get('implementationFeedbackDue').format('YYYY-MM-DD')`
?
Isn't
`metadata.get('implementationFeedbackDue')`
already a string with the same format that you need?

deniak: Oups, I thought I had pushed the missing parenthesis but I didn't. Done now.

I use `Moment` to convert strings like `2018-6-1` to `2018-06-01`. By default, specberus doesn't provide the leading `0`.

lib/transition-checker.js
```diff
@@ -0,0 +1,85 @@
...
+      resolve({ errors: errors, requiresCfE: false });
+    }
+    else if (profile === 'CR') {
+      if (previousVersion.includes('/WD-') || (previousVersion.includes('/CR-') && !isEditorial)) {
+        var octo = new Octokat({
+          token: global.GH_TOKEN
+        });
+        var repo = octo.repos('w3c', 'transitions');
+
+        var shortname = latestVersion.match(new RegExp(/.*\/([^/]+)\/$/))[1];
```
tripu: About this regex: are we sure all *latest version* URLs end with a slash?

lib/transition-checker.js
```diff
@@ -0,0 +1,85 @@
...
+        var directorApproval = 'Transition approved.';
+        var commApproval = 'Draft transition approved.';
+        var directorApprovalFound = false;
+        var commApprovalFound = false;
+
+        repo.issues.fetch(
+          {
+            labels: 'Awaiting publication',
+            state: 'open',
+            per_page: 100 // TODO: get all the issues, not just the first 100
```
tripu: Until we retrieve them all: perhaps add a param [`sort: 'updated'`](https://developer.github.com/v3/issues/#parameters) to make sure we always get the 100 that were most recently updated?

lib/transition-checker.js
deniak: Yes, that's a requirement from pubrules these days.

lib/transition-checker.js
deniak: Yes, good idea!

PR #624, opened 2018-12-04
fix(package): update file-type to version 10.6.0

tripu:
Closes #617


tripu:
[Covered by automatic tests](https://coveralls.io/builds/20411485/source?filename=lib/document-downloader.js#L82)

PR #810, opened 2021-01-28
Limit token method to Github Actions and travis-ci

### LINKED ISSUE #492: Allow sending encrypted un/pw to the TAR endpoint
Bikeshed currently talks to Echidna by sending a TAR file, authed by a W3C user's username and password.  This works fine locally, but if the group wants to allow publishing from their CI infrastructure, it requires revealing the un/pw to the CI.  Travis and other CIs have methods for storing "secrets" that aren't visible to the public, which is good, but not sufficient for this case - the secrets are still visible to anyone with admin rights to the repo, which means we'd be revealing our W3C password to our collaborators in the repo.

Instead, I'd like to just encrypt the username/password pair with a public key, which Echidna holds the private half of, and send that encrypted blob alongside the TAR, rather than doing HTTP auth.  The process would then be:

1. Locally, one of the editors asks Bikeshed to generate the auth token. It takes their username/password, and using its built-in public key, spits out some cyphertext.
2. The editor puts this string into the CI's secrets infrastructure.
3. The CI is then set up to hand Bikeshed the secret, which it sends to Echidna alongside the generated TAR.
4. Echidna uses its private key to decrypt the secret, and verifies that the username/password is correct, and that the username is allowed to publish for this spec (as it does today).
5. Profit.

This effectively bakes the user's credentials into a potentially-stealable token, but the stealing is limited to repo admins, which are already presumed trusted.  The token can't reveal the user's actual un/pw, just, at most, let someone else publish a spec from a repo that the token's user is already authorized to publish from.  (Using the *other* publication method, anyone with repo rights can publish already; this would allow exactly the same thing.)

deniak:
This PR restricts the use of tokens to Github Actions and travis-ci only. This also removes the need to specify a source so we can configure bikeshed to use that method.

Fix #492


views/index.html
```diff
@@ -141,7 +141,7 @@ <h3 id="header-specberus" class="left">Specberus</h3>
     </article>
 
     <footer>
-      <p>Copyright &copy; 2014&ndash;2015 <a href="http://www.w3.org/"><abbr title="World Wide Web Consortium">W3C</abbr></a></p>
+      <p>Copyright &copy; 2021 <a href="http://www.w3.org/"><abbr title="World Wide Web Consortium">W3C</abbr></a></p>
```
vivienlacourba, line 144:
```suggestion
      <p>Copyright &copy; 2021 <a href="https://www.w3.org/"><abbr title="World Wide Web Consortium">W3C</abbr></a></p>
```

lib/ip-checker.js
```diff
@@ -0,0 +1,40 @@
...
+    const ipRangeCheck = require("ip-range-check");
+    const octokit = new Octokit({ auth: global.GH_TOKEN });
+    let allowedIPs = global.TRAVIS_IP; // travis IPs
+    octokit.request('GET /meta').then(res => {
+      if (res.data && res.data.actions) {
+        // GH actions IPs
+        allowedIPs = allowedIPs.concat(res.data.actions);
+      }
+      if (!ipRangeCheck(ip, allowedIPs)) {
+        errors = errors.push('The token method is restricted to Github actions and travis-ci only. If ' +
```
vivienlacourba, line 32:
```suggestion
        errors = errors.push('The token method is restricted to GitHub Actions and Travis CI only. If ' +
```

lib/orchestrator.js
```diff
@@ -302,6 +305,33 @@ Orchestrator.prototype.runUserChecker = function (user, delivererIDs) {
...
+          status: 'failure',
+          errors: errors,
+          history: `This IP ${ip} isn't authorized to use the token method`
+        });
+      }
+    }).catch(function (error) {
+      return new Map({
+        status: 'error',
+        errors: List.of(error.toString()),
+        history: 'An error occurred while running the ip checker.'
```
vivienlacourba, line 329:
```suggestion
        history: 'An error occurred while running the IP checker.'
```

config-dev.js
```diff
@@ -38,3 +38,5 @@ global.LDAP_USER = "foo";
 global.LDAP_PASSWORD = "bar";
 global.LDAP_GROUPS = ["cn=123,ou=groups,dc=w3,dc=org", "cn=456,ou=groups,dc=w3,dc=org", "cn=32113,ou=groups,dc=w3,dc=org"];
 global.SKIP_VALIDATION = true
+
+global.TRAVIS_IP = ["35.224.112.202", "34.122.208.80", "34.74.253.255", "35.192.136.167", "104.154.113.151", "35.202.145.110", "35.192.85.2", "35.222.7.205", "207.254.16.36", "34.66.50.208", "35.192.187.174", "207.254.16.38", "34.73.65.1", "34.66.25.221", "35.192.10.37", "104.196.53.161", "35.202.245.105", "34.74.79.111", "35.192.91.101", "34.73.66.97", "35.237.56.208", "104.198.131.58", "35.227.97.188", "35.229.115.143", "104.196.57.92", "104.154.120.187", "34.122.173.211", "104.197.122.201", "35.231.58.0", "35.188.73.34", "34.74.91.53", "34.68.144.114", "35.193.184.18", "35.227.58.83", "35.237.8.208", "34.66.200.49", "35.196.82.30", "34.74.16.120", "34.73.34.132", "35.196.72.151", "35.193.7.13", "35.184.226.236", "35.185.97.135", "35.196.158.85", "207.254.16.35", "104.196.213.122", "35.188.15.155", "207.254.16.39", "104.154.182.187", "34.66.178.120", "207.254.16.37", "35.196.99.99", "35.237.212.185", "35.193.14.140", "35.188.1.99"]
```
vivienlacourba, line 42: Similarly to what you are doing to automatically retrieve the GitHub servers IP addresses, you could do the same for the Travis CI servers.

From: https://docs.travis-ci.com/user/ip-addresses/

> Note that these ranges can change in the future, and the best way to keep an updated safelist is to use the current A records for the NAT hostname shown above, such as by using dig:
>
> `dig +short nat.travisci.net | sort`
>
> or using a service like dnsjson.com to interact with JSON:
>
> `curl -s https://dnsjson.com/nat.travisci.net/A.json | jq '.results.records|sort'`

The first command that relies on the DNS A records is the fastest one and should also be quite reliable removing the need to store those addresses locally (having a separate cron does not seem mandatory, this could be done real time in the code).

Side note: that parameter should not be part of the config-dev.js file

vivienlacourba (changes requested):

config-dev.js
deniak: addressed in https://github.com/w3c/echidna/pull/810/commits/19028c0674c911458647254698beed59ea8c6fd2

vivienlacourba (approved):

PR #836, opened 2021-04-15
Remove 'third-party-resource-checker' because Specberus is already checking them.

jennyliang220:
Specberus integrates the check in https://github.com/w3c/specberus/pull/1099


deniak (approved):
https://github.com/w3c/echidna/blob/main/allowlist.json can also be deleted

deniak (approved):

PR #1073, opened 2022-09-05
Support for REC with candidate amendments

### LINKED ISSUE #1048: Bug: Auto-Publishing as PER failing
Tried to auto-publish a PER of [Aria in HTML](https://w3c.github.io/html-aria/), but failed:  

Found while trying to publish “echidna.tar” (see the [publication job](https://labs.w3.org/echidna/api/status?id=e14d8a80-7e09-4f39-b9c1-fe3d477f6bc5))

> "Error: Only WD, CR, CRD and Notes are allowed!"

Shouldn't PER be allowed also? 

CC @plehegar.

deniak:
Fix #1048

### DIFF
--- a/lib/transition-checker.js
+++ b/lib/transition-checker.js
@@ -32,10 +32,7 @@ TransitionChecker.check = (profile, latestVersion, previousVersion, isUpdate) =>
     let errors = new List();
 
     if (
-      profile === 'WD' ||
-      profile === 'NOTE' ||
-      profile === 'DNOTE' ||
-      profile === 'CRD' ||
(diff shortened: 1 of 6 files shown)


lib/transition-checker.js
```diff
@@ -32,10 +32,7 @@ TransitionChecker.check = (profile, latestVersion, previousVersion, isUpdate) =>
     let errors = new List();
 
     if (
-      profile === 'WD' ||
-      profile === 'NOTE' ||
-      profile === 'DNOTE' ||
-      profile === 'CRD' ||
+      ['WD', 'CRD', 'WD', 'REC', 'DNOTE', 'NOTE'].includes(profile) ||
```
jennyliang220, line 35: Duplicate `WD`s? @deniak

PR #1111, opened 2023-01-03
Fix spec test

jennyliang220:


test/test.js
```diff
@@ -384,13 +384,18 @@ describe('SpecberusWrapper', () => {
       expect(SpecberusWrapper.validate).to.be.a('function');
     });
 
-    const myDraft = server.getMetadata('navigation-timing-2');
+    let content;
+
+    before(() => {
```
deniak, line 389: why moving this in `before()`?  That hook is supposed to prepare for the test but here, it doesn't seem to be needed. Same comment apply for the tests below

deniak (changes requested):

test/test.js
deniak: @jennyliang220 the tests failed without the `before()` hook because of the specberus wrapper caching. With this cache, mocha was using the same instance of specberus and while it's running in parallel, there are some unwanted side effects due to race conditions. More work needs to be done to make SpecberusWrapper cachable.
I've remove that part and the tests pass again without the hook.

deniak (approved):

test/test.js
jennyliang220: ok @deniak

PR #1125, opened 2023-04-03
Transfer to Es modules

jennyliang220:
resolves https://github.com/w3c/echidna/issues/1099
close #1130


app.js
```diff
@@ -214,7 +221,7 @@ const processRequest = (req, res, isTar) => {
         // eslint-disable-next-line no-console
         if (dryRun) console.log('Dry-run: omitting e-mail notification');
         else {
-          mailer.sendMessage(
+          await sendMessage(
```
deniak, line 224: is there any reason to use `await` here?

config-dev.js
```diff
@@ -1,45 +1,49 @@
...
-  /https?:\/\/w3c\.github\.io/i,
-];
-global.LDAP_URL = 'ldap://localhost:1389';
-global.LDAP_SEARCH_BASE = 'ou=user,dc=example,dc=org';
-// LDAP_BIND_DN must contain the placeholder {{username}}
-global.LDAP_BIND_DN = 'uid={{username}},ou=user,dc=example,dc=org';
-global.GH_TOKEN = '';
-global.GH_DIRECTOR_TEAM_ID = '2797096';
-global.GH_COMM_TEAM_ID = '2794457';
+function setGlobalInfo() {
```
deniak, line 3: This is the dev config. In production, we are using a different config file based from `config.js.example`.

deniak (changes requested):

lib/history.js
```diff
@@ -4,14 +4,17 @@
...
+import { importJSON } from './util.js';
 
-const config = process.env.CONFIG || 'config.js';
+const meta = importJSON('../package.json', import.meta.url);
+const { Stack } = pkg;
+
+// const config = process.env.CONFIG || 'config.js';
 // eslint-disable-next-line import/no-dynamic-require
-require(`../${config}`);
+// import(`../${config}`);
```
deniak, line 17: delete comments

lib/orchestrator.js
```diff
@@ -1,26 +1,24 @@
...
+import IPChecker from './ip-checker.js';
+
+const { List, Map } = pkg;
 
 // Configuration file
-const config = process.env.CONFIG || 'config.js';
+// const config = process.env.CONFIG || 'config.js';
 // eslint-disable-next-line import/no-dynamic-require
-require(`../${config}`);
+// import(`../${config}`);
```
deniak, line 21: delete comments

deniak (changes requested):

jennyliang220:
## [Import a module for its side effects only](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import#import_a_module_for_its_side_effects_only)

```
(async () => {
  if (somethingIsTrue) {
    // import module for side effects
    await import("/modules/my-module.js");
  }
})();

```

This could work for the global variables.

deniak (approved):

PR #1325, opened 2026-06-30
Update specberus-wrapper and tests for Specberus v13

kfranqueiro:
This updates echidna to work with the [recently-released Specberus v13](https://github.com/w3c/specberus/releases/tag/v13.0.0).

Note that the `lib/specberus-wrapper.js` diff will be easier to view with whitespace ignored.

## Changes

- Use async functions and chain off of Specberus' own returned promises instead of needing to create and manage them manually
- Use returned promise instead of listening to `end-all` and `exception` events
- Use `.on` directly on Specberus instance (no more need for `sink`)
- Preserve existing `err` event handler logic in `validate`
- Remove `metadata` event handler (this event has not existed for a long time)
- Remove duplicated instantiation code in `extractMetadata`, instead reusing `getSpecberus`
- Remove completely unused `errors` variable from `extractMetadata`
- Update tests to expect native promises from `SpecberusWrapper` APIs

### DIFF
--- a/lib/specberus-wrapper.js
+++ b/lib/specberus-wrapper.js
@@ -1,16 +1,10 @@
- 
 /**
  * @file A [Specberus](https://github.com/w3c/specberus) wrapper. Validate the compliance of Technical Reports with [publication rules](https://www.w3.org/pubrules/doc).
  */
 
 'use strict';
 
-import Promise from 'promise';
-import pkg from 'immutable';
-import util from 'util';
-import { EventEmitter } from 'events';
-
-const { List, Map } = pkg;
+import { List } from 'immutable';
 
 /**
  * @exports lib/specberus-wrapper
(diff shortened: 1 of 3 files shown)


lib/specberus-wrapper.js
```diff
@@ -19,118 +13,74 @@ const { List, Map } = pkg;
...
-    const sink = new Sink();
-    let errors = new List();
-
-    sink.on('err', (type, data) => {
+SpecberusWrapper.validate = async (url, stateMetadata) => {
+  let errors = new List();
+  let specberusProfile;
+
+  const profile = stateMetadata.get('profile');
+  const patentPolicy = stateMetadata.get('patentPolicy');
```
deniak: This is no longer needed

```suggestion
```

deniak (changes requested):
While checking the tests, I was surprised that the test l.419 was not failing:
```ts
it('should return a metadata property that is a Map', () =>
  expect(content)
    .to.eventually.have.property('metadata')
    .that.is.an.instanceOf(Map));
```

but the type definition says:
```ts
export interface SpecberusResult {
    metadata: Record<string, any>;
    ...
}
```

Looking closely, it appears that [mocha doesn't `await` async describe callbacks](https://github.com/mochajs/mocha/issues/2975).

So the test should be rewritten to add the validation in `before(async () => { content = await ...; })`.

@kfranqueiro I would fix the tests in a separate PR but it might be simpler to just tackle this in that PR since it'll show the error with the `Map`

kfranqueiro:
> Looking closely, it appears that [mocha doesn't `await` async describe callbacks](https://github.com/mochajs/mocha/issues/2975).
> 
> So the test should be rewritten to add the validation in `before(async () => { content = await ...; })`.

Fixing the tests is simpler than that. I'm not sure why there was an `await` there at all; it shouldn't have had one, given that all of the tests in that suite already use chai-as-promised anyway.

I've fixed that test suite but am noticing that others also have that problem so I'll have a closer look.

kfranqueiro:
I transferred my initial fix to the `validate` test suite to a separate branch and addressed the remaining issues with the existing test suite in #1326. (Edit: Only the async-related change to the `validate` test suite has been moved; the change of one specific test case to expect Object instead of Map is specific to this branch.)

I've also reduced the `package.json` diff in this PR (to remove npm's automatic alphabetization of dependencies) to avoid conflicting with that PR, so that this should be able to be synced cleanly against main after that one is merged.

deniak (approved):

PR #1326, opened 2026-07-01
Fix async tests

kfranqueiro:
This fixes several test suites, which Denis pointed out in https://github.com/w3c/echidna/pull/1325#pullrequestreview-4605870546 were not properly verifying due to async/await being used at the suite (`describe`) level.

## Changes

- Resolves suite-level `async` usage by instead directly returning any chai-as-promised invocations (`expect(...).to.eventually....`)
  - The async/await usage was added in #1231, probably in an attempt to fix the deeper issue that this PR now addresses
  - See the "Notice:" section in [chai-as-promised's README](https://github.com/chaijs/chai-as-promised#chai-assertions-for-promises)
- Removes `nav-csserrors` and `nav-csswarnings` test cases which seem to be obsolete as of #1114, which removed CSS validation
  - Also removes the `csserror` and `csswarning` conditions in `test/lib/htmlvalidator.js`, as they don't seem relevant after #1114 (they never evaluate to true), even though that's specifically when they were added
  - CSS validation was presumably the only meaningful difference between these two documents; they were now effectively identical other than the order of a few definition list items and some whitespace
- Addresses CI timeouts in now-fixed tests by adding the usual environment variables to make puppeteer use the runner's preinstalled version of Chrome
  - Verified on my fork [before](https://github.com/kfranqueiro/echidna/actions/runs/28543351095/job/84622416949) vs. [after](https://github.com/kfranqueiro/echidna/actions/runs/28543564896)
- Adds chai-related `@types` packages to make editing tests a bit easier in VSCode/VSCodium (regardless of this project not directly using TypeScript at the moment)


test/lib/htmlvalidator.js
```diff
@@ -32,13 +32,6 @@ validator.get('/nu', (req, res) => {
     },
   };
 
-  if (url.indexOf('csserror') !== -1) {
```
deniak, line 35: the `errors` and `warnings` arrays are no longer needed as well.

test/test.js
```diff
@@ -416,46 +414,10 @@ describe('SpecberusWrapper', () => {
...
-      metadata,
-    );
-
-    it('should return an error property that has no errors', () => {
-      expect(content).that.eventually.has.property('errors');
-      content.then(content => {
-        assert.equal(content.errors.size, 0);
-      });
-    });
+        .that.is.an.instanceOf(Object));
```
deniak: I'm guessing you changed the type assertion to prepare the test for specberus 13

deniak (changes requested):

test/test.js
kfranqueiro: Whoops, yep; I moved this test's change over from the other branch forgetting that I had not only fixed the async part, I also changed the verification condition. I'll move that part back over.

test/lib/htmlvalidator.js
kfranqueiro: Thanks, I'm too used to TypeScript yelling at me about this sort of thing (or eslint, but this repo is effectively running with an empty config)

deniak (approved):

